PR #24, opened 2015-01-19
add applyProbabilities mechanism

LukeLiush:
reformat code for change

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,7 +25,7 @@
   <parent>
     <groupId>org.apache.tika</groupId>
     <artifactId>tika-parent</artifactId>
-    <version>1.7</version>
+    <version>1.8-SNAPSHOT</version>
     <relativePath>tika-parent/pom.xml</relativePath>
   </parent>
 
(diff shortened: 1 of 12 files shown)


LukeLiush:
https://issues.apache.org/jira/browse/TIKA-1517

LukeLiush:
@chrismattmann 
@grossws
Hi, i reformatted the code for change, please kindly let me know if you are able to view the change.

LukeLiush:
@Lukeliush
Test

chrismattmann:
Thanks @LukeLiush - I have an idea here. How about you create a sub-class of MimeTypes called BayesianMimeTypes and then override the methods with what you need? Also no need to put comments with /* [luke] */ etc. in there

LukeLiush:
Hi @chrismattmann , thanks a lot for following this up.
I thought about sublcassing MimeTypes when i was implementing it, but I saw that the MimeTypes in the development/trunk version is defined to be a "final" class like the String class which is not extendable...

I will get rid of the [Luke]....:)

PR #27, opened 2015-02-05
Update RarParser.java

kongxianghe1234:
if you want to detect a file which has chinese-like fileName ? this way you did will be error.
details TIKA-936.

### DIFF
--- a/tika-parsers/src/main/java/org/apache/tika/parser/pkg/RarParser.java
+++ b/tika-parsers/src/main/java/org/apache/tika/parser/pkg/RarParser.java
@@ -86,7 +86,7 @@ public class RarParser extends AbstractParser {
                         subFile = rar.getInputStream(header);
 
                         Metadata entrydata = PackageParser.handleEntryMetadata(
-                                header.getFileNameW().equals("")?header.getFileNameString():header.getFileNameW(),
+                                "".equals(header.getFileNameW())?header.getFileNameString():header.getFileNameW(),
                                 header.getCTime(), header.getMTime(),
                                 header.getFullUnpackSize(),
                                 xhtml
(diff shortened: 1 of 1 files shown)


tika-parsers/src/main/java/org/apache/tika/parser/pkg/RarParser.java
```diff
@@ -86,7 +86,7 @@ public void parse(InputStream stream, ContentHandler handler,
                         subFile = rar.getInputStream(header);
 
                         Metadata entrydata = PackageParser.handleEntryMetadata(
-                                header.getFileNameString(),
+                                header.getFileNameW().equals("")?header.getFileNameString():header.getFileNameW(),
```
uschindler: Use "".equals(header.getFileName()) syntax, because this will cause NullPointerException if the wide file name is not available. Alternatively add a null check.

kongxianghe1234: you said is better, but member variables should not be null , Come to this issue , What i mean is that var header has two methods 1. getFileNameString()    2. getFileNameW()  . Your unit test case just cover iso-8859-1 encoding 
tip:https://github.com/apache/tika/blob/trunk/tika-parsers/src/test/resources/test-documents/test-documents.rar
i have tested for chinese and other language , it didn't work , please 
try to debug for this and you will find you have to check for the charset of the fieName.

uschindler: Hi,
this comment was more about code style: If you use equals to compare 2 things, it is always better to have the part that is static and/or final and for sure not null (in this case: the string literal) first. Then you don't eben have to think about if it could be null or not. If you check your code with PMD it will suggest this to you.

PR #28, opened 2015-02-05
add unit test for grb files, and modify the tika-mimetypes.xml to includ...

LukeLiush:
...e the grb magic and glob.


LukeLiush:
improve the IOUtility class

chrismattmann:
Thanks @LukeLiush

PR #30, opened 2015-02-17
Rollback (TIKA-1354) and update pax-exam to version 4.4.0

hlavki:
Test all bundle tests cases passes


chrismattmann:
Thanks @hlavki I applied your patch!

PR #35, opened 2015-03-17
TIKA-1365: Lower priority for XML starting with comment

mkr:
TIKA-1365: Lower priority for XML starting with comment, allow HTML starting with comment to be detected as text/html


chrismattmann:
Thanks @mkr !

PR #39, opened 2015-03-31
Refactor TIKA-1558. Remove service loading blacklist

tpalsulich:
* Remove all direct service loading logic regarding a blacklist.
* Small changes to CompositeParser logic to make sure subclasses of excluded Parsers are also excluded.
* Added new testing in the tika-core module to test regular and subclass exclusion.

@Gagravarr, can you look this over?


Gagravarr:
To help readability, I'd probably be inclined to have `CompositeParser` have an `isExcluded` method, which does both the `excludeParsers.contains` and `assignableFrom` check

Also, the `ParserExcludeTest` might want to live with the other config related exclusion tests, in one of tika-core/src/test/java/org/apache/tika/config/TikaConfigTest.java or tika-parsers/src/test/java/org/apache/tika/config/TikaParserConfigTest.java

tpalsulich:
Thanks! Done. Is that better?

Gagravarr:
Looks good to me!

PR #41, opened 2015-04-11
add probabilistic mime selection

LukeLiush:
The probabilistic mime selection detector has been re-implemented.
the Bayesian probabilistic selection has been improved by adding more freedom and flexibilities that allow users to specify their own prior and conditional probabilities; the concrete details and basic idea are illustrated in the Tika-1571 and the implementation still follow the main description posted in the https://issues.apache.org/jira/browse/TIKA-1517 but with a minor change on the prior probability which was not modifiable in the original design.
The https://issues.apache.org/jira/browse/TIKA-1535 is resolved according to Prof Mattmann's suggestion.


chrismattmann:
@LukeLiush this is great. Can you open up a small Pull request that cherry picks tika-core/src/main/java/org/apache/tika/mime/MimeTypes.java and that commit? then I can resolve TIKA-1535 with that, and then please update the pull request here for 1517 to not include that part. Also, so this will work out of the box, without a model, correct?

LukeLiush:
Thanks @chrismattmann ;
yes, this one"probabilistic mime selection" in this pull request#41 work out of the box without a model.
Regarding the small pull request, i am not sure i understand it correctly...
Do you want to me create a pull request for the code change I made only in the MimeTypes.java? I actually have made some change in the MimeTypes.java in this pull request, I have got rid of the private access modifier of the 2 methods i.e. List<MimeType> getMimeType(byte[] data) and byte[] readMagicHeader(InputStream stream), so that i can call them from the ProbabilisticMimeDetectionSelector class. This small change seems to only reflect what the tika 1535 might be for.  not sure if you want me to create a pull request like this.

LukeLiush:
@chrismattmann , i have added some code to support the tika facet. As an example making facet calls with this feature, please kindly refer to the unit test in this pull request "ProbabilisticMimeDetectionTestWithTika"; i am also write you another update with this too.

chrismattmann:
Yes please create a new or with just the mike types change

PR #44, opened 2015-04-23
Cbor extension  - set cbor magic priority to 50

LukeLiush:


Gagravarr:
Any chance you could explain why this needs to be 50 not 40?

(A unit test which shows the problem would be ideal!)

LukeLiush:
Thanks @Gagravarr for asking.
please refer to the jira issue in the following link for details. 
https://issues.apache.org/jira/browse/TIKA-1610
I believe cbor file is a prettified json, which might take content as its value , the content can be an entire html or xhtml doc. If xhtml is present in the cbor file, then tika will not even bother to look at those type that has priority lower than the xhtml's priority(i.e. 50); i am assigning 50 to cbor in tika-mime-types.xml to have that included in the magic possible list.

PR #46, opened 2015-04-29
TIKA-1532: DIF Parser and change mime-type from text/dif+xml to application/dif+xml

HyperDunk:
Implementation of DIFParser (GCDM Directory Interchange Format) with unit test and change to mime-type as per discussion in TIKA-1532


Gagravarr:
Generally looks good

One comment though - not sure that DIFContentHandler belongs in core? My hunch is that because it's specific to the DIF parser, it should live in the parsers package under DIF, much as the ODF one does - see under tika-parsers/src/main/java/org/apache/tika/parser/odf/

chrismattmann:
@HyperDunk can you update?

HyperDunk:
Yes. I will do the required changes by EOD today.

HyperDunk:
@Gagravarr & @chrismattmann : Since the DIF content is of XML type I have included DIFParser.java in tika-parsers/src/main/java/org/apache/tika/parser/xml/ and hence I should be moving DIFContentHandler.java to the same. Other option will be to move both DIFParser and DIFContentHandler to tika-parsers/src/main/java/org/apache/tika/parser/dif/ (for which I will need to create a new package).
Please suggest.

Gagravarr:
ODF, iWorks, OOXML, RDF and Atom are all xml-based (and some others too). However, all of those have their own packages, as while they're XML based, they're specific to one particular flavour of XML. So, based on that precedent, I'd say the DIF parser and handler should be in the parsers module under org.apache.tika.parser.dif

HyperDunk:
Yes. That makes sense. I will do the changes and submit.

HyperDunk:
Changes are done, tested and committed. Please review and accept pull request.

chrismattmann:
thanks @HyperDunk !

PR #51, opened 2015-06-18
fix for TIKA-1659 contributed by rami

Rshomali:


Gagravarr:
Would you be able to re-do this without the whitespace changes? And ideally also with a small sample file we could use for a unit test?

Rshomali:
I have submitted another commit to undo the whitespace changes. You can find a sample file to test here:
https://issues.apache.org/jira/secure/attachment/12740398/hacker_news.ipa

PR #58, opened 2015-10-13
lower priority on magic for application/xhtml+xml 

jeremybmerrill:
to avoid misdetecting xhtml-containing emails as XHTML docs.

Emails I have (happy to share if you want) contain XHTML, as one part of a multipart email. Prior to this pull request, the priority on the `application/xhtml+xml` magic detector was 50, equal to the priority on the `message/rfc822` detector. Because of the relative position of the two detectors in `tika-mimetypes.xml`, the emails were incorrectly detected as XHTML documents.

With this PR, by downgrading the priority of `application/xhtml+xml` to 40, the more-sensitive email magic detectors take precedence, causing the emails to be properly detected as `message/rfc822`.

I have not run this thru the govdocs tester or anything other than my own documents, so, full disclosure, this could cause false negative xhtml-detections elsewhere.

### DIFF
--- a/tika-core/src/main/resources/org/apache/tika/mime/tika-mimetypes.xml
+++ b/tika-core/src/main/resources/org/apache/tika/mime/tika-mimetypes.xml
@@ -3937,7 +3937,9 @@
   </mime-type>
 
   <mime-type type="application/xhtml+xml">
-    <magic priority="50">
+    <!-- The magic priority for xhtml+xml needs to be lower than that of -->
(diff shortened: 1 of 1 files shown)


chrismattmann:
thanks @jeremybmerrill can you throw up a JIRA issue for this and link here? I'll get this committed today!

jeremybmerrill:
oh man I forgot. sorry! will do.

jeremybmerrill:
https://issues.apache.org/jira/browse/TIKA-1771

PR #60, opened 2015-10-23
Add parse(file, metadata) method

thammegowda:
The 'file' api is convenient to pass file extension information to the parser (especially when custom-mimetype for a glob pattern is configured)

> P.S. I was trying to activate `GeoParser` on `*.geot` files https://github.com/thammegowda/tika-geo-ner-model and it was helpful.


chrismattmann:
thanks @thammegowda and @Gagravarr

PR #61, opened 2015-10-31
Fix for TIKA-1787 : NamedEntityParser 

thammegowda:
Added NamedEntityParser that supports loading of different NER implementations at runtime.
The default NER implementation based on OpenNLP is supplied.

~~Another implementation based on StanfordCoreNLP is located [here](https://github.com/thammegowda/tika-ner-corenlp/blob/master/src/test/java/edu/usc/cs/ir/tika/ner/NamedEntityParserTest.java) This is GNU GPL 3, So kept separate.~~ See UPDATE 2 below

@chrismattmann This is not  100% complete, here are few TODOs : 
1. The NER implementing class name needs is to be read from tika config if possible/available. Currently relying on Java Properties. Please suggest me on how to resolve [this todo](https://github.com/thammegowda/tika/commit/e96da2bc28d5eef81d034e39eb05099ed5d38ac1#diff-3a416957f7629e40c8fe5e8fd6c577a3R70)
EDIT : ~~2. Looking for a best way to read parsed text from non text streams within the NamedEntityParser (not sure if a parser can read output of previous parsers like html or pdf). Please suggest me on how to resolve [this todo](https://github.com/thammegowda/tika/commit/e96da2bc28d5eef81d034e39eb05099ed5d38ac1#diff-3a416957f7629e40c8fe5e8fd6c577a3R91)~~ Using secondary parser to get text content

UPDATE : 1. [Added Regex Based NER ](https://github.com/thammegowda/tika/commit/ea7871bd4afae7d18e500ffc285e58afd08f5e86). Though this can recognize much more patterns than names, _(I am using it for recognising weapon names and weapon types )_

UPDATE 2 : [Added Core NLP NER with runtime class binding](https://github.com/thammegowda/tika/commit/084985b3612438e9ca7107fecdffd67757d04d10) this one is still using java binding instead of command invocation, because :
  +  The commandline binding is not portable across environments (or need to maintain those many ports)
  +  setup overead in distribted environment like hadoop.

UPDATE 3 : Chaining support : 
   Now we can chain many NER Implementations (OpenNLP, CoreNLP, RegEx) to the NamedEntityParser.

PR #62, opened 2015-11-05
fix for TIKA-1787 contributed by Yueheng He

TaichiHo:
Succeed in building using java 1.8.0_65. 
To see the effect, create a text file like the following. 
```
Good afternoon Rajat Raina, how are you today? Hi, I am Tom Brady. I go to school at Stanford University, which is located in California.
```
Save it as test.ner and feed it to tika. 
```
java -classpath tika-app/target/tika-app-1.12-SNAPSHOT.jar org.apache.tika.cli.TikaCLI -m test.ner
```
The result should look like this
```
Content-Length: 137
Content-Type: application/stanford-ner
LOCATION: [California]
ORGANIZATION: [Stanford University]
PERSON: [Rajat Raina, Tom Brady]
X-Parsed-By: org.apache.tika.parser.DefaultParser
X-Parsed-By: org.apache.tika.parser.stanfordNer.StanfordNerParser
resourceName: test.ner
```


Gagravarr:
On the whole, we try to avoid raw un-prefixed metadata keys. We do have some, mostly older ones or ones from other standards, as you've seen! But ideally we try to use a prefix so that people know which standard to go check to understand what a key means. For example, we prefer `dc:title` to `title`, as the former makes it clearer what it means, the latter is now deprecated

Any chance you could replace keys like `ORGANIZATION` with something more like `prefix:name` based on a well-known external metadata definition? (May end up using multiple prefixes)

TaichiHo:
Thank you for your suggestion. What do you think a good prefix would be like in this case?

The other thing is the way I integrate this might not possible due the GPL license, as is pointed out by https://issues.apache.org/jira/browse/TIKA-1787. I am not sure if I should continue work on this. 

The other issue is using a different approach. Might want to look at that. https://github.com/apache/tika/pull/61

Gagravarr:
License wise, GPL dependencies are a no-go for ASF projects. You could maintain it independently though, but users would need to review the license and install manually if they were able to abide by the GPL restrictions. http://wiki.apache.org/tika/3rd%20party%20parser%20plugins is where we maintain the list of incompatibly licensed plugins

In terms of the prefix, I don't know, you're the expert in the tool rather than me! You'll need to read up the documentation from the tool, and find out from that what it means by the various outputs. If it defines them against a well known external standard, then great! Use that. If not, see if any well known metadata standards cover the same logical things, and map onto those

TaichiHo:
Thanks so much.  I will look into it.

PR #63, opened 2015-11-10
 fix for TIKA-1791 - non hierarchical URI exception

thammegowda:
Improvement : Model is loaded once and NameFinder is reused


thammegowda:
UPDATE : 

+ The fix for non hierarchical URI is done by using URL instead of URI and path string. (Learned that we can have a URL to files inside ZIP archive, but not URI)

While I modified NER model loading code to make above change possible, I also happened to make these changes:

+ The NER model was previously reloaded for every `parse()` call. It now reuses the model by making use of a state variable.
+ The `isAvailable()` function was previously trying to launch an external process for every call to figureout availability of 'lucene-geo-gazeteer' command. This has been changed to use a state variable.
+ The model is loaded on first call to `parse()` or `isAviable()` : via lazy intialization. My tests showed that it is backward compattible. 





CC : @Gagravarr @chrismattmann

PR #69, opened 2016-01-10
Psd parser

doykle:
Stepping through the byte stream to find text layer content ran into two blockers:
+ Undocumented 4 bytes `Txt ` coming before the expected `TEXT` key
+ If the first key consumed is not `Txt TEXT`, the key which IS consumed must be supported in order to get to the next key. The size of the data each key identifies is not easily determined.   

This implementation places the entire block of keyed data into a byte array and scans for the `TEXT` key. 

The dead-end code is commented out, saved for an especially rainy day when we can add support for all the keys and do this properly. (Should I do the right thing and put it on a separate branch?)

References:
PSD File Specification from 2013:
http://www.adobe.com/devnet-apps/photoshop/fileformatashtml/#50577409_pgfId-1031423
Unofficial PSD anatomy lesson from 2014:
https://github.com/layervault/psd.rb/wiki/Anatomy-of-a-PSD-File


doykle:
My apologies for opening this PR. The code is a mess and was meant for testing in our fork.

PR #78, opened 2016-02-26
Tika 1875

prasadns14:
Updated netcdf mime type magic number
File - tika-mimetypes.xml


Gagravarr:
There looks to be an unrelated change in here, and the whitespace/indenting looks inconsistent / messed up. Any chance you could fix that so we can merge?

(Also, we do have a test NetCDF file in the tika parsers test documents area, so if you could add a detection unit test using it too, that'd be great!)

prasadns14:
Made the changes suggested.
Have also add a detection unit test for NetCDF file

PR #82, opened 2016-03-01
Fix for TIKA-1882

mkampasi:
The following mime magic has been added to tika-mimetypes.xml to better detect the below mime-types:

1. **application/vnd.ms-cab-compressed (.cab files)** - pattern "MCSF" in the first 4 bytes
2.  **application/vnd.xara (.xar files)** - pattern "xar!" in the first 4 bytes
3. **application/x-mobipocket-ebook (.mobi files)** - pattern "BOOKMOBI" starting at byte position 60
4. **video/quicktime (.mov files)** - patterns "free" and "wide" seen starting at byte position 4


Gagravarr:
To make it easier to read, would you be able to re-do the first 3 patterns as text rather than hex?

mkampasi:
Please find the requested changes in eb2d06b.

After merging with the latest file, I found that mime magic for *.cab and *.mobi files had already been updated. Hence did not update for *.cab files. However, the offset for the "BOOKMOBI" pattern of *.mobi files was set as 23. During my analysis, I found some files with the pattern at offsets of 58 and 60. Therefore, I updated the mime magic to use a range from 0:60. 
Tested the updated tika-mimetypes.xml file for mp4 and quicktime formats as well.

PR #83, opened 2016-03-01
Fix for TIKA-1881

NamithaGS:
Updated Mime-Magic for 6 mime types:
1. application/postscript :  files begin with pattern "%!PS-Adobe-3.0 EPSF-3.0".
2. application/wordperfect: files begin with pattern "ÿWPC" .
3. image/tiff : updated pattern for "MM.+" for Big endian format.(occur at the beginning of files of tiff mime type)
4. application/rdf+xml : updated pattern "rdf" ( from byte offset 5 to 400) 
5. application/atom+xml : updated pattern "feed" ( from byte offset 5 to 50)
6. application/rss+xml : updated pattern "rss" ( from byte offset 5 to 50)

### DIFF
--- a/tika-core/src/main/resources/org/apache/tika/mime/tika-mimetypes.xml
+++ b/tika-core/src/main/resources/org/apache/tika/mime/tika-mimetypes.xml
@@ -58,9 +58,7 @@
   <mime-type type="application/atom+xml">
     <root-XML localName="feed" namespaceURI="http://purl.org/atom/ns#"/>
     <root-XML localName="feed" namespaceURI="http://www.w3.org/2005/Atom"/>
-    <magic priority="50">			
-       <match value="0x66656564" type="string" offset="5:50"/>  <!-- feed -->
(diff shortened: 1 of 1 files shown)


Gagravarr:
For the Atom, RSS and RDF ones - is the magic required? Doesn't the XML detector get them already via the namespace? And without risk of mis-detecting text files which happen to mention feed or rss or rdf near the start?

For the Postscript one - could you re-do this as text rather than hex, so it's easier to read?

(Others look fine!)

NamithaGS:
For the Atom, RSS and RDF :
This was from the FHT analysis in these files. We found some of these files classified into Octet-Stream, and all these 3 types had the occurrence of the pattern may times in the first 50 bytes or so. I based this purely on the analysis and cannot hence find any information about this on the web. As you mention, your advice is to remove these patterns to be on the safer side, updating the same in this pull request.
For the Postscript one : Changed into readable format.

PR #86, opened 2016-03-03
fix for TIKA-1841 contributed by zetisam

zetisam:


tika-parsers/src/main/java/org/apache/tika/parser/microsoft/HSLFExtractor.java
```diff
@@ -85,113 +84,106 @@ protected void parse(
...
             if (notes != null) {
                 xhtml.startElement("div", "class", "slide-notes");
 
+                
+                
+                // Notes text
                 textRunsToText(xhtml, notes.getTextParagraphs());
+                
+                // Note headers/footers can be extracted, but content will be duplicate
+                // Repeat the notes footer, if set
```
chrismattmann: commented code should be removed.

dstevenson:
Is there a reason this is being held off on? The current output from XSLF extraction is very hard to parse to do anything useful with since it's a flat structure that may or may not contain a notes block.

kkrugler:
Hi @dstevenson - the last activity on this (via [TIKA-1841](https://issues.apache.org/jira/browse/TIKA-1841) was @chrismattmann requesting an update based on his input (this was back in Aug 2016). I think that's why it stalled out.

I'm guessing it wouldn't be very hard to bring the patch into alignment with the current code base, but it's not mergeable currently.

dstevenson:
@zetisam are you able to merge master to resolve conflicts here? Otherwise we can maybe close this and reopen a new one we have access to

zetisam:
I'll see if I can get to it somewhere in the coming days.

PR #91, opened 2016-03-09
TIKA-1508 : Add uniformity to parser parameter configuration - contributed by Thamme Gowda

thammegowda:
1. Added `Configurable` interface.
 This can be used for all services like `Parser`, `Detector` which can take
  configurable parameters.

2. Added `ConfigurableParser` interface which extends `Parser` interface.
   I didn't add new method to existing `Parser` because
    that will break the compatibility.

3. `AbstractParser` extends `ConfigurableParser` and has
  default implementation for configure() contract.
  I think it is safe to do so and it doesn't break anything.
  In addition, all parsers which extend `AbstractParser` can easily
  access config from TikaConfig if they want to

3. Added a TODO to `TikaConfig`,
 after this should allow multiple instances of same parser with
 different runtime configurations.

4. `TikaConfig` is modified to detect if instance can be configured,
  if so, then checks if params are available in XML file, parses the
  params and invokes configure(ctx) method with these params

5. Added `DummyConfigurableParser` that simply copies parameters to
 metadata for the sake of testing

6. Added a sample XML config file for testing.
Added `ConfigurableParserTest` that performs an end to end test of all
the above.


thammegowda:
TODOs:
1.  Add `Map<String,String> getParams()` to the `Configurable` interface 
2. Use `TikaParameterConfigException` a subclass of `TikaException` for config errors.

chrismattmann:
@thammegowda I really like this! Let's finish it in 1.14 (going to move on with 1.13 without it).
Can you:

1. update this to the latest apache:master
2. notify me when that is done?

thammegowda:
@chrismattmann sorry for the delay. 
* updated this PR with apache:master

thammegowda:
There are some comments from @tballison regarding the types on [Jira](https://issues.apache.org/jira/browse/TIKA-1508).

PR #96, opened 2016-04-02
fix for TIKA-1925 contributed by Nilay Chheda

mit2nil:
@chrismattmann Please review the change and let me know they can be contributed back to Tika. 
Issue description: [https://issues.apache.org/jira/browse/TIKA-1925](https://issues.apache.org/jira/browse/TIKA-1925)

### DIFF
--- a/tika-core/src/main/java/org/apache/tika/parser/external/ExternalParser.java
+++ b/tika-core/src/main/java/org/apache/tika/parser/external/ExternalParser.java
@@ -24,12 +24,15 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.Reader;
+import java.util.Arrays;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.Locale;
 
 import org.apache.tika.exception.TikaException;
 import org.apache.tika.io.IOUtils;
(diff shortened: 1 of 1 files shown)


mit2nil:
This was discovered while doing Assignment-2 for class CSCI 599 Spring 16 at USC.

chrismattmann:
@mit2nil wondering why the change here from String[] to ArrayList<String> and what that actually accomplishes?

mit2nil:
@chrismattmann Usage of env variable in unix/mac/linux is such that, it sets some environment variable and then run command. Ref. https://en.wikipedia.org/wiki/Env
On windows, there is no exact alternative to do that. To workaround on windows, we execute 'set' command to set an environment variable and then execute another command in same environment. 
To achieve that, I switched to arraylist so that for Windows, I can delete command string "env" and its argument and pass rest of the commands as is.

chrismattmann:
got it. will test this now @mit2nil

chrismattmann:
I think a better way to achieve this is the following @mit2nil after reviewing the PR:

1. Create a helper function isWindows() and set it to the check that you do for windows
2. In the constructor for the class, check if isWindows(), and then if so, simply use a different command to set it rather than env.

chrismattmann:
Yikes, looking at this a different way I'm actually now unclear how to generically solve this other than to simply override the configuration if you are running on windows for the external parser for EXIF tool.

mit2nil:
May be better solution is to simply remove usage of env variable from config file at
https://github.com/apache/tika/blob/master/tika-core/src/main/resources/org/apache/tika/parser/external/tika-external-parsers.xml and handle {OUTPUT} parameter internally in the code. After all it is just a placeholder to store temporary file path.

chrismattmann:
can you show me what this would look like @mit2nil ?

mit2nil:
Below is the example command as per the TIKA wiki on EXIFTool.
`java -Dtika.config=exif-tika-config.xml -classpath tika-app/target/tika-app-1.9-SNAPSHOT.jar org.apache.tika.cli.TikaCLI -m spaghetti-to-sushi.mp4`

As per my understanding, Tika uses all external parser available (defined in .../tika-external-parsers.xml) for MIME types specified in "exif-tika-config.xml" and run respective command. For EXIFTool, filename "spaghetti-to-sushi.mp4" is propagated from Tika CLI to command line parameter to Exiftool. Similarly, Tika CLI should be able to accommodate a new command line flag indicating creation of temporary file which internally can be done using isWindows() kind of utility function.

chrismattmann:
sorry my point was, if there is an update to make to the code that deals with this let me know. The biggest thing I saw was for me on Mac OSX/Darwin without running the env command in front of exiftool it wouldn't work.

PR #99, opened 2016-04-06
Corrected and Improved

reevapp:
The original code did not work at all, the WebClient was an Instance Variable and not only it was not thread-safe but also it would only work for the very first request (all subsequent requests would include duplicated query parameters and fail).
Some error control was added to allow for troubleshooting, before there would be no clue on the source of a failed request.
It is now possible to specify the User Key.

### DIFF
--- a/tika-translate/src/main/java/org/apache/tika/language/translate/Lingo24Translator.java
+++ b/tika-translate/src/main/java/org/apache/tika/language/translate/Lingo24Translator.java
@@ -26,12 +26,15 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.util.Properties;
 
+import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.cxf.jaxrs.client.WebClient;
 import org.apache.tika.exception.TikaException;
 import org.apache.tika.language.LanguageIdentifier;
 import org.apache.tika.language.LanguageProfile;
+import org.apache.tika.language.translate.Lingo24Translator;
+import org.apache.tika.language.translate.Translator;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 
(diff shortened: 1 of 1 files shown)


chrismattmann:
@reevapp I cannot tell what changes you actually made that weren't formatting. Also I have been using this for a bit and didn't notice the behavior you state - so would love to see what you actually changed so I can review it and test it? Thanks.

tika-translate/src/main/java/org/apache/tika/language/translate/Lingo24Translator.java
```diff
@@ -43,72 +46,78 @@
...
+		if (!this.isAvailable())
+			return text;
+		LanguageIdentifier language = new LanguageIdentifier(
+				new LanguageProfile(text));
+		String sourceLanguage = language.getLanguage();
+		return translate(text, sourceLanguage, targetLanguage);
+	}
+
+	@Override
+	public boolean isAvailable() {
```
reevapp: "isAvailable" became a method so it can check for correct setup at any time, which allows for setting the User-Key, before it was expecting it from the properties file only at instantiation time.

tika-translate/src/main/java/org/apache/tika/language/translate/Lingo24Translator.java
```diff
@@ -43,72 +46,78 @@
...
+		try {
+			ObjectMapper mapper = new ObjectMapper();
+			JsonNode jsonResp = mapper.readTree(responseText.toString());
+			if (jsonResp.findValuesAsText("errors").isEmpty()) {
+				return jsonResp.findValuesAsText("translation").get(0);
+			} else {
+				throw new TikaException(jsonResp.findValue("errors").get(0).asText());
+			}
+		} catch (JsonParseException e) {
+			throw new TikaException("Error requesting translation from '" + sourceLanguage + "' to '" + targetLanguage + "', JSON response from Lingo24 is not well formatted: " + responseText.toString());
```
reevapp: If receiving a non-Json response, the previous implementation would only state that an error occurred, now the message received from the server will be returned

tika-translate/src/main/java/org/apache/tika/language/translate/Lingo24Translator.java
```diff
@@ -43,72 +46,78 @@
...
+		}
+	}
+
+	@Override
+	public String translate(String text, String sourceLanguage,
+			String targetLanguage) throws TikaException, IOException {
+		if (!this.isAvailable()) {
+			return text;
+		}
+		WebClient client = WebClient.create(LINGO24_TRANSLATE_URL_BASE);
```
reevapp: WebClient is not thread-sfe and was declared as instance variable before, after the first request it would actually append query parameters to the already defined parameters for each subsequent request.
For instance, on the third request to the server it would regerate a request with URL like this: https://.....?source=es&source=es&source=es......

tika-translate/src/main/java/org/apache/tika/language/translate/Lingo24Translator.java
```diff
@@ -43,72 +46,78 @@
...
+		} catch (JsonParseException e) {
+			throw new TikaException("Error requesting translation from '" + sourceLanguage + "' to '" + targetLanguage + "', JSON response from Lingo24 is not well formatted: " + responseText.toString());
+		}
+	}
+
+	public String getUserKey() {
+		return userKey;
+	}
+
+	public void setUserKey(String userKey) {
```
reevapp: User-Key can now be dynamically set in runtime

chrismattmann:
@reevapp that helps a lot. Thanks. So, I like these changes. Now back to the other part. You have changed the tabs/formatting, and there is a bunch of extraneous textual changes here that aren't actually code changes. Can you reduce this simply to the relevant lines and keep the tabs (in the form of whitespace) the way they were? I'll be happy to merge it. Also have a look at: http://github.com/apache/tika/#contributing-via-github to get an idea of what to call the PR and how to create a JIRA issue that maps to this. Thanks!

reevapp:
No problem, first time doing pull request.... I will make the changes and submit again.

chrismattmann:
thanks @reevapp

PR #109, opened 2016-04-23
Persona Parser 

shivensaiwal:
I have implemented a persona parser and require a peer review.

### DIFF
--- a/tika-parsers/src/main/java/org/apache/tika/parser/persona/PersonaParser.java
+++ b/tika-parsers/src/main/java/org/apache/tika/parser/persona/PersonaParser.java
@@ -1,3 +1,16 @@
+/*
+ *	Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


tika-parsers/src/main/java/org/apache/tika/parser/persona/PersonaParser.java
```diff
@@ -0,0 +1,123 @@
...
+				
+		      
+		        builder.append(entry.getKey());
+		        builder.append(":");
+		        builder.append(entry.getValue());
+		    
+		        builder.append("\n");
+		    }
+		} catch (FailingHttpStatusCodeException e) {
+			// TODO Auto-generated catch block
```
chrismattmann: remove

tika-parsers/src/main/java/org/apache/tika/parser/persona/PersonaParser.java
```diff
@@ -0,0 +1,123 @@
...
+			e.printStackTrace();
+		} catch (SolrServerException e) {
+			// TODO Auto-generated catch block
```
chrismattmann: remove

chrismattmann:
thanks @shivensaiwal please remove the TODOs and I think this is good to go. Can you start on a page on the Tika Wiki for how to install the PersonaExtractor dependency here: http://wiki.apache.org/tika/PersonaExtraction (you will need to create an account on the Tika wiki).

PR #111, opened 2016-04-27
fix for TIKA-1938 contributed by naegelejd

naegelejd:
Adds HtmlParser support for `<script>` tags within `<head>`.


naegelejd:
I don't know how Tika's release process works but is it possible to get this fix into 1.13? It's a trivial patch but an important fix for Nutch.

PR #119, opened 2016-05-03
Implement Tika NER for Grobid quantities - Team 11 USC

minhptx:
Implement Tika NER for Grobid quantities

### DIFF
--- a/tika-parsers/src/main/java/org/apache/tika/parser/ner/grobid/GrobidNERecogniser.java
+++ b/tika-parsers/src/main/java/org/apache/tika/parser/ner/grobid/GrobidNERecogniser.java
@@ -1,216 +1,123 @@
-package org.apache.tika.parser.ner.grobid;
-
-import org.apache.cxf.jaxrs.client.WebClient;
+import org.apache.http.HttpResponse;
+import org.apache.http.NameValuePair;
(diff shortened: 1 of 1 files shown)


tika-parsers/src/main/java/org/apache/tika/parser/ner/grobid/GrobidNERecogniser.java
```diff
@@ -1,216 +1,123 @@
...
                 }
-                entities.put("MEASUREMENT_NUMBERS",measurementNumberSet);
-                entities.put("MEASUREMENT_UNITS",unitSet); 
-                entities.put("MEASUREMENTS",measurementSet);
-                entities.put("NORMALIZED_MEASUREMENTS",normalizedMeasurementSet);
-                
             }
+            return resultMap;
+        } catch (Exception e) {
+//            e.printStackTrace();
```
chrismattmann: remove commented code

chrismattmann:
there seems to be a lot of orthogonal changes here @minhptx . What exactly is this doing over the current implementation? What does it add?

minhptx:
@chrismattmann sorry for my late response. I think the main difference is different library usage. For httprequest and json parser.

PR #120, opened 2016-05-17
fix for TIKA-1941 contributed by Mark Duske

reevapp:
Class transformed into thread-safe and allows for a Lingo24 User-Key to be dynamically assigned


chrismattmann:
there are a ton of orthogonal changes here I don't get. I'm going to close this. @reevapp

PR #122, opened 2016-05-24
TIKA-1978 Invocation of java.net.URL.equals(Object), which blocks to do domain name resolution, in org.apache.tika.parser.geo.topic.GeoParser.initialize(URL)

lewismc:
This issue addresses https://issues.apache.org/jira/browse/TIKA-1978


chrismattmann:
+1

PR #123, opened 2016-05-26
TIKA-1986 : support for typed parameters from XML configuration

thammegowda:
This is a sub-task of TIKA-1508 (please merge #91 first).

This extends configuration file with a place for specifying parameters and their types.

+ Added a class called `Param`
+ Relying on JAXB annotations to convert between XML and Java objects
+ Test case updated to check for types


tika-core/src/main/java/org/apache/tika/parser/ParseContext.java
```diff
@@ -196,6 +202,39 @@ public SAXParserFactory getSAXParserFactory() {
...
+     */
+    public void setParam(String key, Param<?> value){
+        this.params.put(key, value);
+    }
+
+    /**
+     * Gets the value associated with given parameter
+     * @param key parameter name
+     */
+    public void getParam(String key){
```
tballison: This should actually return something, no?  :)

thammegowda:
@tballison 
UPDATE : 
+ Annotation added : Calling it [@Field](https://github.com/apache/tika/blob/67941a63949280289c1a7b5430a86de0aa3838a9/tika-core/src/main/java/org/apache/tika/config/Field.java) becausse the class `Param` already exist. This annotation can be placed on a Field or a setter method.
+ [ An utility](https://github.com/apache/tika/blob/67941a63949280289c1a7b5430a86de0aa3838a9/tika-core/src/main/java/org/apache/tika/utils/AnnotationUtils.java) is added to initialize the annotated objects! 
+ A Test parser - [DummyParametrizedParser](https://github.com/apache/tika/blob/67941a63949280289c1a7b5430a86de0aa3838a9/tika-core/src/test/java/org/apache/tika/parser/DummyParametrizedParser.java) to [test these] (https://github.com/apache/tika/blob/67941a63949280289c1a7b5430a86de0aa3838a9/tika-core/src/test/java/org/apache/tika/parser/ParametrizedParserTest.java)
+ And more [suit of tests](https://github.com/apache/tika/blob/67941a63949280289c1a7b5430a86de0aa3838a9/tika-core/src/test/java/org/apache/tika/utils/AnnotationUtilsTest.java) on annotations for various cases.

chrismattmann:
+1 LGTM

PR #135, opened 2016-09-28
Fix for TIKA-2099

theobisproject:
A test file is provided. Unit test is missing because I don't know where it should be located


theobisproject:
I provided an additional file as testcase (two tared pdf) and rebased against the current master.
It would be nice if this issue can be resolved soon.

theobisproject:
Closing in favour of a194bc4ba2b771807e93e664245f24e3573a15df

PR #136, opened 2016-09-30
TIKA-2106.  Need to lowercase the output file to match the format passed to tesse…

epugh:
…ract cmd line.


epugh:
This is a bug fix to apply on top of TIKA-2093.

PR #145, opened 2017-01-14
Tika5

ashutoshvsingh:
change snapshot to 1.0.0


chrismattmann:
I'm not exactly sure what this does...and it changes things it shouldn't so...

PR #147, opened 2017-01-17
TIKA-2231: Improved param validation of TesseractOCRConfig.setLanguage()

ham1:
I also improved and added more test cases.


ham1:
I've done a bit more than the bug report.
1. I've treated "" as invalid - partly because `TesseractOCRParser` builds the command line without checking if it's empty
2. anything ending with a + as invalid - not sure if this is enforced by Tesseract but it just seemed right

PR #148, opened 2017-01-18
be more parsimonious wrapping streams

joshbooks:
it looks like a bunch of streams were getting wrapped in BufferedInputStreams just to make extra double sure that mark was supported, but this is not as harmless as it might otherwise seem when you run into big nested package files


joshbooks:
Created a ticket on JIRA, see [TIKA-2244](https://issues.apache.org/jira/browse/TIKA-2244)

PR #150, opened 2017-02-01
Fixed TesseractOCRConfigTest and some TesseractOCRConfig refactoring

ham1:
I created a PR for a small bug in TesseractOCRConfig.java the other week (and it's been merged, thanks).
However, the existing tests don't seem to actually be able to detect errors; i.e. they pass even if the methods reject valid params.

I've re-written them but have some questions:

1. setResize(), the javadoc say "Valid range of values is 100-900." but the code seems to only to accept values to the nearest 100 i.e. 100, 200, ..., 800, 900.
Does anyone know if the current behaviour is intended?

2. setDensity(), the javadoc says 150-1200 but the [ImageMagick docs mention 72-1200](https://www.imagemagick.org/script/command-line-options.php#density)

Which is 'correct'/intended?

Please let me know.


ham1:
Any thoughts or should this be closed?

PR #160, opened 2017-03-27
TIKA-2312 proposition of exposing ID3TagsAndAudio

lukaszozimek:
[Issue Link]
https://issues.apache.org/jira/browse/TIKA-2312


Gagravarr:
There seems to be some unrelated whitespace changes in your commit - any chance you could remove those?

Otherwise, commit looks fine, but can I ask what you're trying to get at? Perhaps we might want to expose those id3 fields as well as standard through Tika is all!

lukaszozimek:
@Gagravarr 
Currently I am writing custom Mp3 Parser which allow user to import Metadata from Radio Automation System which name is AudioVault. This tool is used by professional broadcasters.
This system has custom metadata structure for all supported audio files. During the research I recognised that it add system specific metadata which depends on occurrence of ID3v2 tags in any version. To be straight if ID3v2 occurring this system add their specific metadata on top of file if not on the bottom

PR #166, opened 2017-04-07
fix for TIKA-2044 contributed by wko27

wko27:


tika-parsers/src/main/java/org/apache/tika/parser/mbox/MboxParser.java
```diff
@@ -67,23 +63,73 @@
     private static final Pattern EMAIL_ADDRESS_PATTERN = Pattern.compile("<(.*@.*)>");
 
     private static final String EMAIL_HEADER_METADATA_PREFIX = "MboxParser-";
-    private static final String EMAIL_FROMLINE_METADATA = EMAIL_HEADER_METADATA_PREFIX + "from";
-    private final Map<Integer, Metadata> trackingMetadata = new HashMap<Integer, Metadata>();
-    private boolean tracking = false;
+    static final String EMAIL_FROMLINE_METADATA = EMAIL_HEADER_METADATA_PREFIX + "from";
+    private static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("EEE, d MMM yyyy HH:mm:ss Z", Locale.US);
```
grossws: Why are you sure that `MboxParser` would never be called from different threads?

tika-parsers/src/main/java/org/apache/tika/parser/mbox/MboxParser.java
```diff
@@ -188,27 +251,33 @@ private void saveHeaderInMetadata(Metadata metadata, String curLine) {
...
             metadata.add(Metadata.CONTENT_TYPE, headerContent);
             metadata.set(TikaCoreProperties.FORMAT, headerContent);
-        } else {
+            break;
+        default:
             metadata.add(EMAIL_HEADER_METADATA_PREFIX + headerTag, headerContent);
         }
     }
-}
+}
```
grossws: Don't strip EOF whitespace, please

wko27: Oops, re-added!

tika-parsers/src/main/java/org/apache/tika/parser/mbox/MboxParser.java
wko27: Hm, apologies.  TBH, I'd rather switch usage to Java 8's thread-safe DateTimeFormatter or joda-time, but ... seems like Tika is still intended for Java 7 and I'd rather not introduce joda-time as a dependency.  Reverted ...

wko27:
Ping!  Any updates? :)

PR #171, opened 2017-04-25
TIKA-2340 cleanup of direct dependencies / fix locale-dependent test case

user667:
This is an effort of dependency cleanup (with the help of mvn dependency:analyze and others). Not sure if this is acceptable without TIKA ticket number (or as such), I'll be glad to adjust this PR.

We've been updating to a recent (1.6-1.14) tika version lately and found a delta of +40libs/+20MB and went on to investigate. Turns out tika-parsers relies on a lot of transitive dependencies.

These commits try to specify (some) direct dependencies for libraries that are used in imports of parser classes. Additionally one of the test cases failed on my local setup with a non-default-English locale.

### DIFF
--- a/tika-parsers/src/test/java/org/apache/tika/parser/microsoft/ooxml/OOXMLParserTest.java
+++ b/tika-parsers/src/test/java/org/apache/tika/parser/microsoft/ooxml/OOXMLParserTest.java
@@ -1467,6 +1467,9 @@ public class OOXMLParserTest extends TikaTest {
 
     @Test
     public void testXLSBVarious() throws Exception {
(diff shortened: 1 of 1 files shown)


grossws:
Thank you for contribution, @user667.

Could you please squash these commits excluding test fix and open separate PR for test fix?

Feel free use [TIKA-2340](https://issues.apache.org/jira/browse/TIKA-2340) as tracking issue. I'll merge your patch after squashing.

user667:
BTW: This is just a small percentage of what should be added as direct dependencies. Unfortunately I don't know most of the parsers and the libraries they use and therefore cannot provide the rest of the list (see ```mvn dependency:analyze```).

PR #173, opened 2017-04-28
Fix for TIKA-2347 Adds underline extraction from word documents

stuarthendren:
Extracts underline for both doc and docx and assigns tag <u>.
Given lowest nesting among style tags.
Adds tests using testWORD_various.doc and testWord_various.docx
Updates affected output in other WordParserTests.


darkdreamingdan:
Could you also add strikethrough support?  It's just the same thnig but using the <strike> xhtml element.  We have our own branch for this code but it would be good to unify our PRs

dameikle:
Merged into master now with included strikethrough support for docx. Thanks @stuarthendren!

PR #174, opened 2017-04-28
TIKA-2332 Add SNOMED codes to server output

oniofchaos:
https://issues.apache.org/jira/browse/TIKA-2332


chrismattmann:
LGTM! thanks @oniofchaos

PR #176, opened 2017-05-03
fix for TIKA-2352 contributed by pascal.essiembre

essiembre:


tballison:
Wow, that was fast.  Thank you!!!

PR #181, opened 2017-05-17
TrueTypeParser Close Open Fonts

icirellik:
The TrueTypeParser opens fonts during parsing, however it never closes
these resources. This causes a slow file handle leak that causes
problems when processing large amounts of fonts.

This change wraps the font resource in a try-finally block to ensure
that the resource is closed after parsing has occurred.

### DIFF
--- a/tika-parsers/src/main/java/org/apache/tika/parser/font/TrueTypeParser.java
+++ b/tika-parsers/src/main/java/org/apache/tika/parser/font/TrueTypeParser.java
@@ -59,49 +59,55 @@ public class TrueTypeParser extends AbstractParser {
             Metadata metadata, ParseContext context)
             throws IOException, SAXException, TikaException {
         TikaInputStream tis = TikaInputStream.cast(stream);
-        
+
         // Ask FontBox to parse the file for us
-        TrueTypeFont font;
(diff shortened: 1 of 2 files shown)


Gagravarr:
I'm not sure if it'd be easy to do, but if you could come up with a unit test for this (failing on Tika trunk/master, passing with your patch applied), that'd be great to ensure we don't accidentally break it again at some point in the future!

icirellik:
I added a unit test that is able to reproduce the leak. I wouldn't necessarily suggest merging the test into master. It only works (tests anything) on unix systems, it will run on other systems.

As you said the issue is hard to test, being that is caused by the `RAFDataStream` in the FontBox library never being closed and there is no easy was to programmatically see which files the JVM has open.

tballison:
@Gagravarr sorry for not waiting for you to take this.  I wanted to make sure to get it in before 1.15.

@icirellik thank you for this PR!  I followed your recommendation to remove the unit test.  We can add it back later if fellow devs want it.  Again, many thanks!

tballison:
I opened https://issues.apache.org/jira/browse/TIKA-2370 to track this on our JIRA.

PR #183, opened 2017-06-02
fix for TIKA-2385 contributed by pmweiss5

pmweiss:


dameikle:
Thank for the patch. Unfortunately the test doesn't work for me

dameikle:
Sorry, I didn't mean to close this, not enough morning coffee :) Thanks for the PR. Was looking to include this into the code base but the test image requires the default ImageMagick security limits (on Ubuntu at least) to be lifted. @pmweiss - do you have a smaller test image that would meet the limits? If not, I can create one.

dameikle:
I've added a smaller image and adapted the Python check to check for the dependencies as well to ensure the test can run on machines that have Python but not all of the dependencies. Thanks again @pmweiss

PR #184, opened 2017-06-15
TIKA-2393 Fix Sentiment parser model URLs

thammegowda:
Sentiment parser test models URLs are pointed to a specific commit in git instead of HEAD of the master.


PeterAlfredLee:
The problem of wrong urls has been fix in #330 . Maybe we can close this PR and [TIKA-2393](https://issues.apache.org/jira/browse/TIKA-2393)

PR #185, opened 2017-06-15
TikaFix ---Contribution ny jerni-zu393

jerni-zu393:
I have added the magic bits for three files (*.keystore/.jks , *.apk , *.aac). It can be detect the file types even the files should not have "."extensions .

I have attached here below the sample files for testing purpose .

[keystore.tar.gz](https://github.com/apache/tika/files/1077097/keystore.tar.gz)

[apktest.zip](https://github.com/apache/tika/files/1077108/apktest.zip)
[aactest.zip](https://github.com/apache/tika/files/1077115/aactest.zip)

### DIFF
--- a/tika-core/src/main/resources/org/apache/tika/mime/tika-mimetypes.xml
+++ b/tika-core/src/main/resources/org/apache/tika/mime/tika-mimetypes.xml
@@ -37,6 +37,13 @@
      you can here, then provide a Custom Detector for the rest
 -->
 <mime-info>
+<!--TIKAFIX by jerni-zu393 : Magicbits for keystore file is copied from MimeUtil- magic.mime to detect .keystore/.jks files -->
+  <mime-type type="application/x-java-keystore" >
+  <acronym>keystore</acronym>
+  <magic priority="50">
+  <match value="0xfeedfeed" type="string" offset="0"/>	
+  </magic>
+  </mime-type>	
 
   <mime-type type="application/activemessage"/>
   <mime-type type="application/andrew-inset">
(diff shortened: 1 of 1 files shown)


jerni-zu393:
Any updates ?

tballison:
@Gagravarr any objections?  Would be useful to open a ticket on our JIRA to track to this.  Also would be helpful to add unit tests with files that you've provided.  Thank you!

Gagravarr:
The keystore one should probably go further down the file, so it's in alphabetical order as the others are

For APC files, do we know if they always store the entries in that specific order? Or if it might change? Currently `ZipContainerDetector` only requires `AndroidManifest.xml` to be present, do we want to mirror that or keep your wider list of files required?

The AAC file magic with ID3 at the front doesn't necessarily look right to me - won't most MP3 files with ID3 tags in first incorrectly match on this too?

tika-core/src/main/resources/org/apache/tika/mime/tika-mimetypes.xml
```diff
@@ -4694,6 +4720,18 @@
   <mime-type type="audio/vnd.vmx.cvsd"/>
   <mime-type type="audio/vorbis-config"/>
   <mime-type type="audio/x-aac">
+<!--TIKAFIX by jerni-zu393: magic bits for "aac" file is added here-->
+  <magic priority="50">
+  <match value="P" type="string" offset="2">
+  <match value="libfaac" type="string" offset="11"/>
+  </match>
+  </magic>
```
Savar: where is this detection coming from? I cannot find it the same in the `file` util nor looking around.. my AAC file starts with FFF1 which is kind of indicated in https://www.garykessler.net/library/file_sigs.html as an example

PR #194, opened 2017-07-14
SUPPORTED_TYPES is immutable

onealj:
In case SUPPORTED_TYPES is ever changed back to a set containing more than one value, a reminder to use an immutable container such as `Collections.unmodifiableSet(set)`.

This is a continuation of #193.


chrismattmann:
thanks

PR #200, opened 2017-08-24
TIKA-2447 reduce memory consumption of PSDParser

bjrke:


Gagravarr:
Thanks for this! Unfortunately it breaks the unit tests as it stands, because stream.skip may not skip the whole amount... I've committed a slightly modified version

bjrke:
Thanks for this fast apply of that patch. You should also close the jira ticket.

bjrke:
I've tested the code. To protect us against too large files we crop them.
This now causes the file to unlimeted skip, because it can't skip behind the cropped end of the file.
This is caused by the implementation of skipFully which tries to skip if even there is nothing to skip.
Unlike read, skip will not return -1 if EOF is reached!

Gagravarr:
Any thoughts on how we should patch skipFully to handle it? Maybe check if skip returns 0, then call a read to see if it's -1?

bjrke:
there is a bug report in poi: https://bz.apache.org/bugzilla/show_bug.cgi?id=61294

bjrke:
and it is fixed with https://github.com/apache/poi/commit/c7db66a30dfb6cbbd5812ff3ae4c90ed2d9b9a27

tballison:
@Gagravarr thank you for taking this!  @bjrke thank you for opening this and submitting a PR!  I'm just back from vacation...sorry for not contributing, is there anything that needs to be done on this?

bjrke:
no, only release of a new version of poi and tika :)

PR #201, opened 2017-08-25
directly compare stderr to empty string in testRedirectionOfStreams to obtain more meaningful messages if test fails

boegel:
While installing Tika 1.16, I was running into this cryptic test failure:

```
java.lang.AssertionError: expected:<0> but was:<42>
```

At first, I was a bit confused, since `42` is clearly a superior result compared to `0`... [1]

After doing some digging, it turns out that I had `$_JAVA_OPTIONS` defined, which results in a message like this to stderr every time `java` is called:

```
Picked up _JAVA_OPTIONS: ...
```

I had to change the failing test to directly compare the stderr output with the empty string rather than checking the length of the strings to figure out what was wrong, so I propose to change this.

[1] https://en.wikipedia.org/wiki/42_(number)#The_Hitchhiker.27s_Guide_to_the_Galaxy

### DIFF
--- a/tika-batch/src/test/java/org/apache/tika/batch/fs/BatchProcessTest.java
+++ b/tika-batch/src/test/java/org/apache/tika/batch/fs/BatchProcessTest.java
@@ -232,7 +232,7 @@ public class BatchProcessTest extends FSBatchTestBase {
         assertEquals(1, countChildren(outputDir));
         assertContains("System.out", streamStrings.getOutString());
         assertContains("System.err", streamStrings.getOutString());
-        assertEquals(0, streamStrings.getErrString().length());
+        assertEquals("", streamStrings.getErrString());
 
     }
 


tballison:
Sorry about that, and thank you for figuring out what went wrong.  I completely agree on the superiority of 42.  Does your PR actually fix your problem, though? If you're getting `Picked up...` on stderr...that doesn't equal "".

boegel:
It doesn't fix my problem, but it makes the cause of the failing test very clear (as opposed to `0 != 42` as output from the failing test).

If not `$_JAVA_OPTIONS` is not defined, the `Picked up ...` messages is not printed to `stderr`, and the test passes.

So, maybe you want to undefine `$_JAVA_OPTIONS` in the test to avoid that the test result is determined by the environment?

tballison:
Ha, ok.  It is uglier than I'd like to undefine `$_JAVA_OPTIONS` in the child process.  I modified the test to allow `""` or a string starting with "Picked up"...a kluge, I admit, but...let me know if this fixes your problem.

PR #202, opened 2017-08-28
Convert EncryptedPowerPointFileException to EncryptedDocumentException

mattcg:
This avoids the ambiguous error `org.apache.tika.exception.TikaException: Unexpected RuntimeException from org.apache.tika.parser.microsoft.OfficeParser`.


tballison:
Added unit test.  Thank you for this!

PR #205, opened 2017-09-01
TIKA-2455: flag the containing multipart type

mattcg:
Flag the type of bodies contained within a multipart body.

This allows alternative bodies to be distinguished from attachments, for example, and ensures that information on this structure is not lost in the output from Tika.


tballison:
This looks great!  Would you be able to add a unit test?

mattcg:
Thanks. Yes, I will. I'm also going to add the boundary value.

What d'you think of the names?

tballison:
I like the names.  I wonder if we should add them as Properties to o.a.t.metadata.Message in tika-core?

mattcg:
@tballison updated with a test and also:

1) only store the multipart subtype (the entire type is redundant);
2) store the root multipart type when parsing the fields of the main message.

tballison:
Thank you!

PR #206, opened 2017-09-08
1.6

yaseminaladi:
Hello,

I have a document which file format is UDF(UYAP Dosya Format). When I read the mimetype from bytearray of this document, tika jar finds mimetype as application/zip. What could I do so that tika jar could find mimetype correctly,mimetype is text/udf.

Can you help me, please?


Gagravarr:
For usage questions, please post on the mailing list - https://lists.apache.org/list.html?user@tika.apache.org - or StackOverflow. If you think you've found a bug, please either post on the mailing list, or open a bug in JIRA - https://issues.apache.org/jira/projects/TIKA

PR #210, opened 2017-10-10
fix for TIKA-2475 contributed by seanstory

seanstory:


tballison:
Thank you, @seanstory!

PR #211, opened 2017-10-11
[TIKA-1788] RFC822Parser: provide email attachment filenames when available

AarjavP:
https://issues.apache.org/jira/browse/TIKA-1788 old case but had the solution in it. Not sure why it never made it in.

There is other info available in `MaximalBodyDescriptor`. However I am not sure how they would map to fields in `Metadata` or if most files would even have that info*.

One piece of info that I would like to make available is either the whole `Content-Disposition` line which I am not sure how to get, or flags for `attachment`/`inline` that are almost always available with filename. Looking at the source code, it is parsed but hidden away behind a `private` field.😢

### DIFF
--- a/tika-parsers/src/main/java/org/apache/tika/parser/mail/MailContentHandler.java
+++ b/tika-parsers/src/main/java/org/apache/tika/parser/mail/MailContentHandler.java
@@ -27,6 +27,8 @@ import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.Locale;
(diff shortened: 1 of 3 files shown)


AarjavP:
the second commit 9653e77 also provides `Content-Disposition` in the metadata of the embedded object

tballison:
@AarjavP , I very much appreciate this PR.  I regret that I haven't been able to review it carefully yet, but I look forward to doing so over the next week or so (I hope).  Thank you!

tballison:
Thank you!

PR #212, opened 2017-10-13
 NetCDF-Java Library

med-ali-bannour:
NetCDF-Java Library with the version 4.5.5 is old, i think it should be updated to new version because this library made a problem for me when actually try to migrate Guava library.


lewismc:
Hi @med-ali-bannour please scope out #218

PR #214, opened 2017-12-29
TIKA-2479 - Handle empty cells in XLSX

glb:
The POI parser doesn't emit `cell` events for empty cells (by design),
but they do provide some sample code for how to handle sparse events.
This commit updates `XSSFExcelExtractorDecorator` to emit empty `tr`
and `td` elements for the missing cells.

See also: https://bz.apache.org/bugzilla/show_bug.cgi?id=57775
See also: https://bz.apache.org/bugzilla/show_bug.cgi?id=61937


dameikle:
It looks like Nick Burch hasn't seen this and has added a different fix for this in master.  Does this do what you were expecting?

PR #221, opened 2018-02-21
Fix for TIKA-2581 contributed by ewanmellor.

ewanmellor:
TesseractOCRParserTest.testOCROutputsHOCR fails with Tesseract 4.0.

With 3.x, the output is <span>Happy</span> but with 4.0 the output is
<span><strong>Happy</strong></span>.  Both these seem reasonable to me,
so update the test to accept either of them.

### DIFF
--- a/tika-core/src/test/java/org/apache/tika/TikaTest.java
+++ b/tika-core/src/test/java/org/apache/tika/TikaTest.java
@@ -107,6 +107,11 @@ public abstract class TikaTest {
         assertTrue(needle + " not found in:\n" + haystack, haystack.contains(needle));
     }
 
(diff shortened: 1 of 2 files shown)


Gagravarr:
I wonder if it's be better to either strip the `<strong>` tags out before comparing, or just check for `Happy</` instead?

ewanmellor:
@Gagravarr Both of those options would work, but I don't see how either of them are any better.

epugh:
At this point, does it make sense to support Tesseract3 when running tests?    Maybe update the documentation https://cwiki.apache.org/confluence/display/TIKA/TikaOCR that the output format is slightly different?

PR #231, opened 2018-03-28
Update forbiddenapis to version 2.5 and remove commons-io hack

uschindler:
Hi,
as TIKA is my Maven test project for new versions of forbiddenapis and the local test build already succeeded (before my release), here is a PR that updates to new version. It also removes the commons-io hack in the parent pom.


tbpalsulich:
Thanks!

PR #232, opened 2018-04-02
Fix for TIKA-2624 contributed by ewanmellor.

ewanmellor:
Change AbstractPDF2XHTML.doOCROnCurrentPage to use the same DPI value
(PDFParserConfig.ocrDPI) for both the PDF rendering and the image metadata.

Previously, the PDF was being rendered using ocrImageScale (default 2.0 ==
144dpi) and then putting ocrDPI (default 300) in the image metadata.  Having
these two things be independent makes no sense, and is surely going to
confuse Tesseract when the image metadata does not match the data.

This change means that ocrDPI drives both values, and ocrImageScale is
removed.  This also switches from PDFRenderer.renderImage to
PDFRenderer.renderImageWithDPI, but that's just a stub to make it clearer
what's going on.

This change will have the side-effect that the temporary images between the
PDF rendering and Tesseract will be 4x larger (144dpi to 300dpi).  This will
have a memory and temporary disk space impact, but it will ensure that the
whole pipeline uses 300dpi by default.  People who have memory constraints
will need to reduce ocrDPI and make the corresponding changes on the
Tesseract side.

### DIFF
--- a/tika-parsers/src/main/java/org/apache/tika/parser/pdf/AbstractPDF2XHTML.java
+++ b/tika-parsers/src/main/java/org/apache/tika/parser/pdf/AbstractPDF2XHTML.java
@@ -326,12 +326,13 @@ class AbstractPDF2XHTML extends PDFTextStripper {
         TemporaryResources tmp = new TemporaryResources();
         try {
 
-            BufferedImage image = renderer.renderImage(pageIndex, config.getOcrImageScale(), config.getOcrImageType());
+            int dpi = config.getOcrDPI();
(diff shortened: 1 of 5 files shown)


epugh:
Interesting patch, do you have any examples of this being an issue that you can share?   I never really thought about having them be the same.

ewanmellor:
I don't have examples that I can share, sorry.  This work was for a proprietary product, and I am no longer working on it.

I remember it making an absolutely enormous difference in OCR quality though (with Tesseract 4).

PR #234, opened 2018-04-24
TIKA-2636 ENVI Header metadata fields can span more than one line

lewismc:
This issue addresses https://issues.apache.org/jira/browse/TIKA-2636 and also provides augmented unit tests and a new test resource.
Additionally, this PR addresses [invocation of the EnviHeaderParser via the tika-server](https://s.apache.org/Is7G) by adding the Parser discovery to src/main/resources/META-INF


tika-parsers/src/main/java/org/apache/tika/parser/envi/EnviHeaderParser.java
```diff
@@ -62,27 +72,54 @@ public void parse(InputStream stream, ContentHandler handler,
         if (tikaConfig == null) {
             tikaConfig = TikaConfig.getDefaultConfig();
         }
-        try (AutoDetectReader reader = new AutoDetectReader(
```
Gagravarr: Is there a reason why you removed the try / close auto-handling here for your patch

lewismc: Hi @Gagravarr thanks for looking at the patch. I shifted the AutoDetectReader to global status such that the object can be accessed from the newly introduced ```readLines``` method. 
If you prefer, I suppose we could just pass the AutoDetectReader object to the readLines method call such that that method signature would be ```readLines(AutoDetectReader)```.
Right now I attempt to catch ```(IOException | TikaException e)``` and introduced manually closing the AutoDetectReader.

In my latest push, I've encapsulted paragraph construction into a single method.

Gagravarr: I think it might be good to pass the reader to the `readLines` method, and friends - I'm a little concerned about the thread safety of the current code

lewismc: ack @Gagravarr I've addressed this now.

PR #235, opened 2018-05-02
TIKA-2639 Update freedesktop.org shared-mime-info-spec hyperlink in MimeTypesReader.java

lewismc:
This issue addresses https://issues.apache.org/jira/browse/TIKA-2639


Gagravarr:
Good spot, thanks!

PR #236, opened 2018-05-23
TIKA-2648 : detect interpreted server-side scripting languages

GerardBouchar:
mime detection based on resource name used to detect
the mime-type of "http://example.com/test.php" as being "text/x-php"
whereas given such an URL, the file extension doesn't give
us any information about the mime type that will be returned
by the server


GerardBouchar:
@tballison So, have there been objections ?

tballison:
Sorry for my delay.  I'm still thinking about this.  I've been a bit busy on some other things, but I haven't forgotten!

GerardBouchar:
@tballison anything new on this ?

PR #238, opened 2018-05-25
TIKA-2100 extract content language from html lang attribute

GerardBouchar:
The [recommended way](https://www.w3.org/International/questions/qa-html-language-declarations) of declaring the language of an HTML document is to specify it in the `lang` attribute of the `<html>` tag.

Tika currently not only ignores this attribute, but also removes it from its SAX output, making it unavailable to client applications.

This PR adds the value of the lang attribute to the documents metadata (under the existing key `Metadata.CONTENT_LANGUAGE`) and also makes it available in the SAX output.

### DIFF
--- a/tika-core/src/main/java/org/apache/tika/sax/XHTMLContentHandler.java
+++ b/tika-core/src/main/java/org/apache/tika/sax/XHTMLContentHandler.java
@@ -138,7 +138,12 @@ public class XHTMLContentHandler extends SafeContentHandler {
             
             // Call directly, so we don't go through our startElement(), which will
             // ignore these elements.
-            super.startElement(XHTML, "html", "html", EMPTY_ATTRIBUTES);
+            AttributesImpl htmlAttrs = new AttributesImpl();
+            String lang = metadata.get(Metadata.CONTENT_LANGUAGE);
+            if (lang != null) {
+                htmlAttrs.addAttribute("", "lang", "lang", "CDATA", lang);
+            }
+            super.startElement(XHTML, "html", "html", htmlAttrs);
             newline();
             super.startElement(XHTML, "head", "head", EMPTY_ATTRIBUTES);
             newline();
--- a/tika-parsers/src/main/java/org/apache/tika/parser/html/HtmlHandler.java
+++ b/tika-parsers/src/main/java/org/apache/tika/parser/html/HtmlHandler.java
@@ -16,20 +16,6 @@
  */
 package org.apache.tika.parser.html;
 
-import java.io.ByteArrayInputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.net.MalformedURLException;
-import java.net.URL;
-import java.nio.charset.StandardCharsets;
-import java.util.Arrays;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Locale;
-import java.util.Set;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
 import org.apache.tika.extractor.EmbeddedDocumentExtractor;
 import org.apache.tika.extractor.EmbeddedDocumentUtil;
 import org.apache.tika.metadata.HTML;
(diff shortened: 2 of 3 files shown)


tika-core/src/main/java/org/apache/tika/sax/XHTMLContentHandler.java
```diff
@@ -138,7 +138,12 @@ private void lazyStartHead() throws SAXException {
             
             // Call directly, so we don't go through our startElement(), which will
             // ignore these elements.
-            super.startElement(XHTML, "html", "html", EMPTY_ATTRIBUTES);
+            AttributesImpl htmlAttrs = new AttributesImpl();
```
tballison: Thank you for the patch.  This feels fiddly to me.  Let's leave XHTMLContentHandler alone.

tika-parsers/src/main/java/org/apache/tika/parser/html/HtmlHandler.java
```diff
@@ -16,20 +16,6 @@
  */
 package org.apache.tika.parser.html;
 
-import java.io.ByteArrayInputStream;
```
tballison: If you could turn off your auto-correct for imports or align your imports with ours so that we don't have all these changes, that'd be great!

GerardBouchar: Yes, I saw that too late, sorry :) fixing it now.

tika-core/src/main/java/org/apache/tika/sax/XHTMLContentHandler.java
GerardBouchar: But how can we fix [TIKA-2100](https://issues.apache.org/jira/browse/TIKA-2100) without touching XHTMLContentHandler ?

tballison: Sorry...I'm so narrowly focused only on my own typical use case - extract text without structure, put the lang in Metadata()...

To confirm, you have a need to include the 'lang=' in the xhtml output from Tika?  Are there other common attributes of HTML that we should also include?

GerardBouchar:
I reverted the automatic changes to the import statements (there are unused imports, though).
But I don't see how to add the language information to the output HTML without fiddling with XHTMLContentHandler...

tika-core/src/main/java/org/apache/tika/sax/XHTMLContentHandler.java
GerardBouchar: Hmm, did you mean attributes of the `<html>` tag, or html attributes in general ?

tballison: Sorry, right, attributes of `<html>`.

chrismattmann: let's include :lang in the HTML attributes as well +1.

chrismattmann:
this LGTM - @tballison are we good to commit this?

chrismattmann:
@tballison I hear you and am open to alternatives. What is a better way to do this? I think missing the lang attribute is a pretty bad thing and have seen it in the past. It feels like HTMLParser as a parser should contribute it (and I don't think you object to that) perhaps via Metadata.set and then what should we do to propagate it?

tballison:
If :lang is special, we should treat it specially :).  If there are other
attrs that go on the html entity, we can add them later? Onward!

tballison:
So... go forth!

Side note: It’d be fun to see counts for elements w a lang attr in our
corpus.

tika-core/src/main/java/org/apache/tika/sax/XHTMLContentHandler.java
GerardBouchar: So the answer is no, I do not need other attributes of the `<html>` tag.

GerardBouchar:
So, do I still need to change something ?

tballison:
You’re good. I’m away from my keyboard today but will be back tomorrow.

PR #242, opened 2018-07-12
Rework the standard-compliant charset detector

GerardBouchar:
This commit contains all my latest work on StandardHtmlEncodingDetector.

 * Improves performance
 * Improves standard compliance
 * Adds tests
 * Fixes bugs (the stream was not reset after detection)

Standard compliance is still not perfect, as some parts, such as page-reloading, do not fit well in TIKA.

I am open to any remarks or suggestions concerning the code and am ready to rework it if necessary.

For a comparison to other encoding detectors, see: https://issues.apache.org/jira/browse/TIKA-2673?focusedCommentId=16541307&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16541307

Using this detector, we achieved <1% misdetected charsets in a random crawl segment.


tika-parsers/src/main/java/org/apache/tika/parser/html/charsetdetector/PreScanner.java
```diff
@@ -0,0 +1,268 @@
...
+                stream.reset();
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private void skipUntil(byte... expected) throws IOException {
+        while (!expect(expected)) {
+            if (stream.skip(1) == 0) return;
```
tballison: Depending on the stream type, you can't actually rely on the return value of `skip`.  I've learned the hard way to follow commons-io's `skipFully` and use `read` instead.

tika-parsers/src/main/java/org/apache/tika/parser/html/charsetdetector/StandardHtmlEncodingDetector.java
```diff
@@ -0,0 +1,100 @@
...
+     * A composite encoding detector chaining a {@link StandardHtmlEncodingDetector}
+     * (that may return null) and a {@link StandardIcu4JEncodingDetector} (that always return a value)
+     * This full thus always returns an encoding, and still works very well with data coming
+     * from the web.
+     */
+    public static final EncodingDetector FULL_DETECTOR = new CompositeEncodingDetector(asList(
+            new StandardHtmlEncodingDetector(),
+            new StandardIcu4JEncodingDetector()
+    ));
+    private static final int META_TAG_BUFFER_SIZE = 8192;
```
tballison: Please add the instance-level `markLimit` like we do in HtmlEncodingDetector as a \@Field so that users can configure a different markLimit.

GerardBouchar:
> This is a check-then-set pattern that should be in a synchronized block, right? I may be missing multithreaded guarantees elsewhere in your design... I'm happy with a single static initialization rather than lazy loading.

Good catch, thank you :+1: Charset initialization can be costly, especially if there are custom CharsetProvider loaded, so I chose lazy loading. But if you think this is not necessary, then we can go back to a static initialization.

GerardBouchar:
> What is the benefit of moving from a config file to hard coded loading? I somewhat prefer the config file, but defer to you if there's a good reason.

I didn't really like the custom file format (TSV with comments) with a custom java parser just for it. Plus I wanted to be able to add ReplacementCharset and XUserDefinedCharset without registering them for the whole program.

GerardBouchar:
> When you call PreScanner, you've already set mark() on the parent stream (before wrapping it in the BoundedInputStream). In other words, you assume that the stream sent in to detect already supports mark(). Do you need to wrap this separately?

`EncodingDetector` [guarantees](https://github.com/apache/tika/blob/master/tika-core/src/main/java/org/apache/tika/detect/EncodingDetector.java#L42-L46) that the stream supports `mark()`, and requires the detector to `reset()` it before returning.

tika-parsers/src/main/java/org/apache/tika/parser/html/charsetdetector/StandardHtmlEncodingDetector.java
GerardBouchar: I did this in https://github.com/apache/tika/pull/242/commits/e7cda261f6aa2eb0102a8ffb53e454a4f3655bdd

However, this makes the detector mutable, and makes things more complex. Couldn't we set  the `markLimit` in a constructor parameter, in order to keep the detector immutable ?

PR #243, opened 2018-07-16
Fix for TIKA-2683 contributed by karanjeets

karanjeets:
@kkrugler - Can you please share your thoughts?


tika-parsers/src/main/java/org/apache/tika/parser/html/BoilerpipeContentHandler.java
```diff
@@ -17,11 +17,9 @@
 package org.apache.tika.parser.html;
 
 import java.io.Writer;
-import java.util.ArrayList;
-import java.util.BitSet;
-import java.util.List;
-import java.util.Locale;
+import java.util.*;
```
kkrugler: In general we try to avoid importing `.*` (other than `Assert.*`), though I see a few of these have slipped into the code in other places.

tika-parsers/src/main/java/org/apache/tika/parser/html/BoilerpipeContentHandler.java
```diff
@@ -230,14 +229,20 @@ public void endDocument() throws SAXException {
...
-                            if (validCharacterRuns.get(curCharsIndex)) {
+                            // https://issues.apache.org/jira/projects/TIKA/issues/TIKA-2683
+                            // Allow exempted characters to be written
+                            if (isValidCharacterRun ||
+                                    (chars.length == 1 && whitelistCharSet.contains(chars[0]))) {
                                 delegate.characters(chars, 0, chars.length);
 
                                 // https://issues.apache.org/jira/browse/TIKA-961
-                                if (!Character.isWhitespace(chars[chars.length - 1])) {
+                                if (isValidCharacterRun && i == element.getCharacters().size() - 1
```
kkrugler: I'd move this block up one level - I don't believe it depends on the previous test (`isValidCharacterRun || (chars.length == 1...`) evaluating to true, right?

kkrugler: Could you explain why we now want to check for `i == element.getCharacters().size() - 1`, where that wasn't previously done by the patch for TIKA-961?

kkrugler:
If it was possible to have a smaller, generic HTML file for the test then that would be better. I'm almost certain Apache projects are OK with [CCA Share Alike](https://en.wikipedia.org/wiki/Wikipedia:Text_of_Creative_Commons_Attribution-ShareAlike_3.0_Unported_License) but I'd have to double-check.

tika-parsers/src/main/java/org/apache/tika/parser/html/BoilerpipeContentHandler.java
karanjeets: Sure, I will make the changes.

tika-parsers/src/main/java/org/apache/tika/parser/html/BoilerpipeContentHandler.java
karanjeets: That's correct. I will move this one level up. 👍

karanjeets: It seems `element.getCharacters()` returns a section which corresponds to a tag in the HTML web page. Inside that there are sub-sections created by char length or other heuristics based on the BoilerPipe algorithm. 

This check ensures that we are detecting/adding whitespace characters at the end of section and not in between. The current behavior doesn't take care of this and you may see improper new-line characters between the text (Example - https://en.wikipedia.org/wiki/Blobfish)

karanjeets:
@kkrugler - Thanks a lot for your quick review. I will update the PR with suggested changes and will try to create a smaller/generic HTML file for the test.

karanjeets:
@kkrugler - Updated the PR. Please review.

Honestly, the changes from TIKA-961 seems a hack to me. Ideally, there should be a `ContentHandler` which look at the respective HTML tags and append new-line characters accordingly, similar to `XHTMLContentHandler` logic but without HTML tags. Do you know if anything like this already exist in Tika?

karanjeets:
@kkrugler - I see. Well, I hope this PR resolves some of the whitespace issues. Thanks for reviewing the code changes.

kkrugler:
Thanks @karanjeets!

chrismattmann:
thanks for the review and commit @kkrugler and @karanjeets! 👍

PR #249, opened 2018-09-17
Update forbiddenApis to version 2.6

uschindler:
This mainly adds support for Java 11 class files/signatures and silences the commons-io warnings for modules that don't use commons-io.


tballison:
Thank you, @uschindler !

PR #250, opened 2018-09-21
fix for TIKA-2731 contributed by @jkakavas

jkakavas:
As part of the changes introduced in 1.19 `determineMaxEntityExpansions` needs to read the `jdk.xml.entityExpansionLimit` System Property in order to overwrite the default value of 20, if it is set.
This is however by reading all System Properties with `System#getProperties()` and attempting to find the relevant key in the properties Object. The issue with this approach is that `System#getProperties()` requires:
```
java.util.PropertyPermission "*", "read,write"
```
which is an overly permissive one to allow for the given use case.

A more sane approach, following the least privilege design principal would be to use `System.getProperty()` for the specific property that only requires
```
java.util.PropertyPermission "jdk.xml.entityExpansionLimit", "read"
```


tballison:
Duh...sorry and thank you!

PR #255, opened 2018-10-30
TIKA-2630: Wrong height and width metadata for JPEG images

dameikle:
- Added extraction of image height/width from ExifSubIFDDirectory for compressed images
- Include directory name as key qualifier for Exif directories to avoid clashes


dameikle:
Hey @tballison - given the key name clashes for the Exif metadata I am proposing to add the directory name as the qualifier, hence the request for a review.

I was tempted to do this for all directories to make it clean but worry about downstream code that rely on the current values in 1.x stream.

I also thought about not doing it, but without doing it for at least Exif, we will continue to give the wrong value here without some logic to have a key hierarchy in CopyUnknownFieldsHandler.

dameikle:
@tballison Thanks for looking over it - my main worry was changing the behaviour. There is a fixed structure that one should expect with data included in them only if written, so we could make them static property prefixes.  I think the directory name is safe in this instance given how the format works and agree the prefix route - its easy enough to make this fixed against those directories.  I'll go with that.

tballison:
Sorry, @dameikle, I should have addressed the change in behavior...the actual reason you asked for a second pair of eyes. 😆 I agree that changes in behavior are bad.  IIUC, though, we'd be fixing what we're currently doing, which is over-writing info, right?

If we did something like this in branch_1x:
`    if (directory instanceof ExifDirectoryBase) {
        metadata.set(directory.getName() + ":" + tag.getTagName(), tag.getDescription());
        metadata.set(tag.getTagName(), tag.getDescription());
    } else {
        metadata.set(EXIF_ROOT + ":" + tag.getTagName(), tag.getDescription());
        metadata.set(tag.getTagName(), tag.getDescription());
    }`

That would maintain the same current (wrong) over-writing behavior, and introduce new tag names.  I have no idea what an appropriate prefix would be for EXIF_ROOT...something static and documented and appropriate.

dameikle:
I was thinking something similar but stopped as it is not always going to Exif metadata that is processed by the Extractor. I am beginning to think it might be the right thing to actually fix the overwriting, unless you completely disagree.

saitho:
What's the status on this? :)

tballison:
Unless there are objections, let's put this in 1.23?

dameikle:
@tballison - I agree, let's go with this one

PR #259, opened 2018-11-14
TIKA-2783: Add unit tests for org.apache.tika.mime.HexCoDec

JohnLBergqvist:
Hi, 

I ran JaCoCo over the `tika-core` module, and found that the `org.apache.tika.mime.HexCoDec` class was missing unit test coverage. 

I've written some tests for this class, with the help of [Diffblue](http://www.diffblue.com) [Cover](https://www.diffblue.com/s/Diffblue-Cover-Datasheet.pdf). The majority of the class is now covered as a result of these tests.

Hopefully these tests should help you detect regressions caused by future code changes.


JohnBergqvist:
@kkrugler Many thanks.

JohnBergqvist:
@kkrugler Many thanks. The first test in the file has a `length` of `1`, which causes the exception to be thrown.

PR #262, opened 2019-01-16
Update mp2t glob pattern

mcalcote:
Adding the glob pattern based on the info here: https://www.w3.org/TR/mse-byte-stream-format-mp2t/

### DIFF
--- a/tika-core/src/main/resources/org/apache/tika/mime/tika-mimetypes.xml
+++ b/tika-core/src/main/resources/org/apache/tika/mime/tika-mimetypes.xml
@@ -7102,7 +7102,9 @@
 
   <mime-type type="video/mp1s"/>
   <mime-type type="video/mp2p"/>
-  <mime-type type="video/mp2t"/>
+  <mime-type type="video/mp2t">
+    <glob pattern="*.ts"/>
+  </mime-type>
 
   <mime-type type="video/mp4">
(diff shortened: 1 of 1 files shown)


mcalcote:
No one seems to care about this change ¯\_(ツ)_/¯

nick-burch-flec:
Sorry, looks like this one fell through the cracks!

Do you know of / can you create a very short MPEG-2 TS file that we could use for unit testing with this? (Ideally minimal size, and under the Apache License v2 or similar)

Also, a quick read of the linked spec suggests we might be able to do some mime magic detection based on the initialization segment. [Wikipedia has](https://en.wikipedia.org/wiki/List_of_file_signatures) some vague looking suggestions for the format too, but I'm not sure how accurate / false-positive-prone those are. Is that something you have any experience with to help us develop/understand?

slovdahl:
@nick-burch-flec Very late to the party but here's a 2632 byte TS file with a H264 video stream that I created using an image I took myself and `ffmpeg -framerate 24 -i img%d.jpg -c:v libx264 -pix_fmt yuv420p -crf 51 output.ts`. Feel free to use however you want. I had to wrap it in a ZIP file because GitHub does not allow uploading `.ts` files.

[ts-file.zip](https://github.com/user-attachments/files/20796330/ts-file.zip)

PR #268, opened 2019-03-21
Update slf4j to 1.8.0-beta4

dadoonet:
Because of [CVE-2018-8088](https://ossindex.sonatype.org/vuln/d33d3123-eac8-43d1-a5b1-4ebb82c88b77), we need to upgrade `slf4j` to `1.8.0-beta4`.

Hope this helps.

PR #269, opened 2019-03-21
Update soap-api to 1.4.0 and stax-ex to 1.8.1

dadoonet:
When maven enforcer is running, we are getting some conflicts with those dependencies.

* `stax-ex` is used by `jaxb-runtime` and `cxf-rt-rs-client`
* `soap-api` is used by `cxf-rt-rs-client` only but this one brings 2 different versions on the same lib

Related to the issue I found while working on #268


dadoonet:
I believe that we can close this one?

PR #275, opened 2019-06-20
Switch to using Files.createTempFile

coheigea:
According to the Javadoc for File.createTempFile:

The Files.createTempFile method provides an alternative method to create an empty file in the temporary-file directory. Files created by that method may have more restrictive access permissions to files created by this method and so may be more suited to security-sensitive applications.


VVATOR (approved):
It`s a really good idea to upgrade this

THausherr:
Thanks, I committed your changes into 1.* and 2.* .

PR #278, opened 2019-08-09
TIKA-2830 add heif mimetype support

makepanic:
fixes [TIKA-2830](https://issues.apache.org/jira/browse/TIKA-2830)

I had to bump `com.drewnoakes.metadata-extractor` to `2.12.0` because they [added heif support](https://github.com/drewnoakes/metadata-extractor/releases/tag/2.12.0).

I'm unsure if the mimetypes are correct like that because the sequence types have no glob.
If I add the same glob to each sequence, the xml parser throws an error.
The [spec](https://nokiatech.github.io/heif/technical.html) says that the same extension  can be used for different mime types:

Most of the parser code is copied from the webp parser implementation.


tballison:
@makepanic I'm sorry this took forever.  We had to do some unpleasant shimming to upgrade drewnoakes' metadata extractor.  We've done this now, and this _should_ just work now.  THANK YOU!

PR #281, opened 2019-08-29
TIKA-2931

epugh:
Here is a PR that fixes sending information information to System.err, so you can return clean JSON when using tika-app on the commandline.  

There are some extract commits that are from merging my tika project, and not sure how to remov them.  Maybe just a cherrypick?


tballison:
I'm good with this (once squashed, which I can do at merge time).

I suspect this code was added before we had logging.  I wonder if we should use LOG.info() instead, and change the unit test to check for the existence/size of the output file.

PR #284, opened 2019-10-02
Fix for TIKA-2954 contributed by chriszappia

chriszappia:
See https://issues.apache.org/jira/browse/TIKA-2954


tballison:
Thank you!

PR #298, opened 2019-11-26
fix for TIKA-3001 contributed by tandara0

tandara0:


tandara:
Prevent the POIFSFileSystem from closing InputStream insides by wrapping CloseShieldInputStream.

PR #299, opened 2019-11-28
fix for TIKA-3003 contributed by cesarsotovalero

cesarsotovalero:
I'm making this pull reques because I noticed that dependency `org.jsoup:jsoup:1.12.1` is declared in module `tika-parsers` to prevent from having a vulnerable version from edu.ucar:grib. However, this dependency is not used and, therefore, it can be removed to make the ` pom` clearer and the dependency tree of this module less complex.

In addition, dependency `net.sf.ehcache:ehcache-core`, induced transitively from `edu.ucar:cdm:4.5.5`, is not used and can be excluded safely. Notice that the size of the `jar` of `ehcache-core` is around 1.3MB, thus removing it has a positive impact on the size of module `tika-parsers`.

### DIFF
--- a/tika-parsers/pom.xml
+++ b/tika-parsers/pom.xml
@@ -540,13 +540,6 @@
       <version>5.5.0</version>
     </dependency>
 
-    <!-- grib's current jsoup is vulnerable to xss
-         exclude and import a more modern version TIKA-2561-->
-    <dependency>
-      <groupId>org.jsoup</groupId>
-      <artifactId>jsoup</artifactId>
-      <version>1.12.1</version>
-    </dependency>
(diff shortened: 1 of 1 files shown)


trejkaz:
Looks good to me, who hits the button?

tballison:
If you want to fix the conflicts, that'd be great...or, this is small enough, I can take it.  Let me know.  Thank you!

tballison:
As a side note, we've locked `master` for archival use only now.  We're using `main` for the dev branch Tika 2.0.0, and `branch_1x` for the dev branch for Tika 1.25, etc.

PR #300, opened 2019-12-07
TIKA-2224 - OneNote parser

nddipiazza:
# OneNote parser support for tika

TODO:

- [ ] package private classes that don't need to be public
- [ ] see if we can also parse .onetoc files.
- [ ] apache licensing headers in source. 
- [ ] add optional configuration to the parse context so that you can customize how it parses.
- [ ] currently all revisions are always crawled. and it doesn't work when you try to crawl just the latest revision. 
- [ ] stop returning a big map of everything in it when walking tree. we don't need it.
- [ ] add embedded content handlers to parse the embedded files.
- [ ] get more example files and add them to the unit test. Specifically from more versions of OneNote besides just 2013.


tballison:
Need to review statics to make sure this parser will be thread safe.

PR #302, opened 2019-12-12
Tika 2224 one note2016 test

nddipiazza:
add a onenote 2016 test case
add a onenote 2016 embedded word doc test case


tballison:
@nddipiazza is this overcome by events, or would this still be a useful addition?

nddipiazza:
i think we already merged this in the 1x branch.

PR #307, opened 2020-01-02
Addendum to Tika 2224 - Add support for OneNote 2007 and earlier

nddipiazza:
OneNote versions before OneNote 2010 do not have a published OpenSpec document, and the older formats are drastically incompatible with the later OpenSpecs. 

Therefore, we resort to scraping out useful ASCII and UTF16LE strings using a similar algorithm used by the GNU "strings" program.


tballison:
On Twitter, Den Delimarsky recommended checking out a tool to interrogate old OneNote files: https://twitter.com/DennisCode/status/1234217442076848128?s=20

nddipiazza:
@tballison  interesting. yes the solution in place currently for one note 2007 is way too slow for production. we need a better solution. 

I will use this tool and see how far I get.

PR #310, opened 2020-02-05
TIKA-3037 Update docs for Tika Server

epugh:
introduce the Docker image, bump some versions, link to the right place on the wiki.

I changed the versions to all be 1.24, but maybe everything needs to be 2.0-SNAPSHOT??


epugh:
@tballison thoughts on these changes?

PR #316, opened 2020-04-06
Tika 2368 renable cve checks

epugh:
From a discussion on the mailing list re: the `ossindex-maven-plugin` plugin, discovered the open ticket TIKA-2368 which is about a long list of dependencies.

As far as I can tell we don't use them, though the unit test completely mocking things up makes things hard to tell.  Here is a stab at exclusions that make `mvn test` happy, and the `ossindex-maven-plugin` happy as well..

### DIFF
--- a/tika-nlp/pom.xml
+++ b/tika-nlp/pom.xml
@@ -33,7 +33,7 @@
 
   <name>Apache Tika Natural Language Processing</name>
   <url>http://maven.apache.org</url>
-  
+
 
   <dependencies>
     <dependency>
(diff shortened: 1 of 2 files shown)


THausherr:
I think this PR can be closed because it's obsolete, the module structure has been changed, master is now main. I did some changes in "Apache Tika age recogniser" that aren't the same as yours, but may be in the same spirit. There are now only 4 security risks left.  The best would be to release a new version `age-predictor-api` with all dependencies updated. But I doubt this will happen if this was part of a university project and everybody involved has moved on.

PR #317, opened 2020-04-13
fix for TIKA-3089 contributed by pvanderweerd

pweerd:
Wrapping text in pre-tags instead of p-tags will preserving formatting much better

### DIFF
--- a/tika-parsers/src/main/java/org/apache/tika/parser/csv/TextAndCSVParser.java
+++ b/tika-parsers/src/main/java/org/apache/tika/parser/csv/TextAndCSVParser.java
@@ -231,14 +231,14 @@ public class TextAndCSVParser extends AbstractEncodingDetectorParser {
 
     private static void handleText(Reader reader, XHTMLContentHandler xhtml)
             throws SAXException, IOException {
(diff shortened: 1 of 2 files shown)


tika-parsers/src/main/java/org/apache/tika/parser/csv/TextAndCSVParser.java
```diff
@@ -306,7 +306,6 @@ private CSVParams getOverride(Metadata metadata) {
 
         String delimiterString = mediaType.getParameters().get(DELIMITER);
         if (delimiterString == null) {
-            return new CSVParams(mediaType, charset);
```
tballison, line 309: Why remove this?

pweerd: This removal was unintentional. Apologies for not double checking the diff.
Shall I create a new merge request?

KranthiGV: You can revert it and commit the changes to your branch. It'd automatically show up in this PR.

PR #325, opened 2020-07-10
TIKA-3131 -- swap default values of averageCharTolerance and spacingT…

clarkperkins:
…olerance to match PDFBox defaults


tballison:
Thank you @clarkperkins !

PR #326, opened 2020-07-14
TIKA-3133 - writeLimit and maxEmbeddedResources for recursive parsing - add header

nddipiazza:
see https://issues.apache.org/jira/browse/TIKA-3133
and https://issues.apache.org/jira/browse/TIKA-3126

this will add new parameters to `rmeta` rest endpoint

`writeLimit` - max number of characters to store; if < 0, the handler will store all characters
`maxEmbeddedResources` - number of embedded resources that will be parsed. if < 0, it will handle unlimited embedded resources.

This will make it so we can control how many embedded docs will be parsed in a call to rmeta, and how many bytes will be written to the body.


nddipiazza:
@tballison i think this can be merged now. I disassociated it with TIKA-3126 so that this PR can be 100% focused on not hard coding those values.

PR #327, opened 2020-07-15
fix for TIKA-3134 contributed by tothd

tothd91:


tballison:
@tothd91 thank you for opening this!  It looks like there are quite a few changes that are white-space only.  Would it be possible to update so that the diff includes only logic differences?  Thank you!

tothd91:
Hello @tballison  i did it. I hope it's ok now.

PR #332, opened 2020-07-18
Fix can't del tmp file in windows

PeterAlfredLee:
TestCase `org.apache.tika.image.HeifParserTest.testSimple` failed in windows because `TemporaryResources.close()` sometimes fail to delete tmp file.

We can make it `deleteOnExit` as it's only a temp file, as Windows will eventually delete them from time to time even through `deleteOnExit` failed.

### DIFF
--- a/tika-core/src/main/java/org/apache/tika/io/TemporaryResources.java
+++ b/tika-core/src/main/java/org/apache/tika/io/TemporaryResources.java
@@ -24,6 +24,8 @@ import java.nio.file.Path;
 import java.util.LinkedList;
 
 import org.apache.tika.exception.TikaException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Utility class for tracking and ultimately closing or otherwise disposing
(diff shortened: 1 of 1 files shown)


THausherr:
Isn't this moot?
https://issues.apache.org/jira/browse/TIKA-3135

PeterAlfredLee:
Hi @THausherr , sorry for the late reply.
I think the fix in [TIKA-3135](https://issues.apache.org/jira/browse/TIKA-3135) is trying to avoid occupying the file, therefore we can close it(mostly on Windows).
My fix in this PR is trying to make the file deleteOnExit in exception handle, and they will be eventually deleted. Considering they are temp files, I think this fix is reasonable - we do not need to stop the process if exceptions are thrown when deleting temp files.

THausherr:
I agree that it shouldn't stop the process. Suggestion: output a log message, because the cause is usually a programming oversight, so that it can be reported and fixed.

PeterAlfredLee:
> Suggestion: also output a log message, because the cause is usually a programming oversight, so that it can be reported and fixed.

Just pushed the logging part. :)

PR #334, opened 2020-07-30
Tika-3141 : add empty environment variable handle

PeterAlfredLee:
Trying to fix Tika-3141 with a empty string check in `TikaConfig`

### DIFF
--- a/tika-core/src/main/java/org/apache/tika/config/TikaConfig.java
+++ b/tika-core/src/main/java/org/apache/tika/config/TikaConfig.java
@@ -249,11 +249,11 @@ public class TikaConfig {
     public TikaConfig() throws TikaException, IOException {
 
         String config = System.getProperty("tika.config");
-        if (config == null) {
+        if (config == null || config.trim().equals("")) {
             config = System.getenv("TIKA_CONFIG");
         }
 
-        if (config == null) {
+        if (config == null || config.trim().equals("")) {
             this.serviceLoader = new ServiceLoader();
             this.mimeTypes = getDefaultMimeTypes(getContextClassLoader());
             this.encodingDetector = getDefaultEncodingDetector(serviceLoader);


tika-core/src/main/java/org/apache/tika/config/TikaConfig.java
```diff
@@ -249,11 +249,11 @@ public TikaConfig(ClassLoader loader)
     public TikaConfig() throws TikaException, IOException {
 
         String config = System.getProperty("tika.config");
-        if (config == null) {
+        if (config == null || config.trim().equals("")) {
```
keithrbennett, line 252: Could you use `StringUtils.blank(config)` instead? This would be more concise, and we are using StringUtils methods in other parts of the code as well. (See https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isBlank-java.lang.CharSequence-).

PeterAlfredLee: It seems the `StringUtils` is part of `Apache Commons Lang`, which is not a dependecy of `tika core`. Should I add it in the `pom.xml`?

PeterAlfredLee: And I think we should add a log message here for `config` that is not null but empty. WDYT?

keithrbennett: > It seems the `StringUtils` is part of `Apache Commons Lang`, which is not a dependecy of `tika core`. Should I add it in the `pom.xml`?

Since it is a dependency of other Tika modules:

```
% grep "commons-lang" **/pom.xml
tika-bundle/pom.xml:              commons-lang3|
tika-dl/pom.xml:          <artifactId>commons-lang3</artifactId>
tika-dl/pom.xml:          <artifactId>commons-lang3</artifactId>
tika-dl/pom.xml:          <groupId>commons-lang</groupId>
tika-dl/pom.xml:          <artifactId>commons-lang</artifactId>
tika-eval/pom.xml:            <artifactId>commons-lang3</artifactId>
tika-nlp/pom.xml:          <groupId>commons-lang</groupId>
tika-nlp/pom.xml:          <artifactId>commons-lang</artifactId>
tika-parsers/pom.xml:      <artifactId>commons-lang3</artifactId>
tika-parsers/pom.xml:          <artifactId>commons-lang3</artifactId>
tika-parsers/pom.xml:          <artifactId>commons-lang3</artifactId>
tika-server/pom.xml:      <artifactId>commons-lang3</artifactId>
```

...if you don't hear from anyone soon II would suggest adding it to tika-core and let the person authorizing the pull request merge make that call. I suspect it's ok, because it's used elsewhere in Tika.

PeterAlfredLee: Sure. Maybe we should have it discussed in mailing list.

tballison: Let's keep commons-lang3 out of core for now.  If fellow devs want to add it in 2.0, let's open a separate ticket.

PR #335, opened 2020-08-01
TIKA-307 truncated zip

jendabenda:
Fix for https://issues.apache.org/jira/browse/TIKA-307


tballison:
Fixed in `main`.  Thank you!

PR #342, opened 2020-08-22
Modify TikaCLITest

PeterAlfredLee:
1.remove unnecessary import
2.reset outContent and errContent if they are not empty,prevent previous TikaCLI.main run output left.
3.add two test case
4.modify previous test case,use method getParamOutContent to run TikaCLI.main and get output string.


PeterAlfredLee:
Update :  5.simplify some code in method testExtract, testExtractTgz, testExtractInlineImages.

PR #348, opened 2020-08-29
Refactor some code use method Math.max

PeterAlfredLee:


tika-core/src/main/java/org/apache/tika/io/CountingInputStream.java
```diff
@@ -56,7 +56,7 @@ public CountingInputStream(InputStream in) {
     @Override
     public int read(byte[] b) throws IOException {
         int found = super.read(b);
-        this.count += (found >= 0) ? found : 0;
```
kkrugler, line 59: For this read(), and the following one, I wouldn't use `Math.max()`. It's really a test to see if the `super.read()` returned -1 or an actual read amount, so better would be something like `if (found != -1) { this.count += found; }`

PeterAlfredLee: Good point. Just pushed a new commit. :)

kkrugler:
Thanks Peter!

PR #353, opened 2020-09-09
Fix test fail in SAS7BDATParserTest caused by default language is not english

PeterAlfredLee:
Use short months in default language to test.


PeterAlfredLee:
This is another fix implemention like PR [#331](https://github.com/apache/tika/pull/331), so that we do not need to modify the jvm args like #331 did.

Only eithor one of these 2 PRs need to be merged and I will close the other one.

PR #357, opened 2020-09-14
Modify TikaInputStream

PeterAlfredLee:
After we copy all data in `InputStream` to `file` and use new `InputStream` open from `file`, we should update `position` and `mark`


tballison:
@PeterAlfredLee can you figure out why this hasn't caused problems before?  Is it because we typically spool to disk before reading anything?

PeterAlfredLee:
@tballison 
>why this hasn't caused problems before

Because method `getPath` use original InputStream instead of current TikaInputStream when copy data to a temporary file.
I think this is also a problem because it makes variable `position` may not be real position.
Just fix this with PR #363 

WDYT :)

PR #137, opened 2016-10-25
TIKA-1343 Create a Tika Translator implementation that uses JoshuaDecoder

lewismc:
This issue addresses https://issues.apache.org/jira/browse/TIKA-1343 and is a cleaner PR than the one I just screwed up over at #112

PR #364, opened 2020-09-25
Fix TIKA-3196

PeterAlfredLee:
When reading a zip archive entry with STORED and Data Descriptor, a UnsupportedZipFeatureException would be thrown. We can save the number of entries we have already read, reset the stream, and open the ZipArchieInputStream again with Data Descriptor allowed. Then we can finish reading the rest of the entries.

1. I set a limit of 100MB using variable `MARK_LIMIT`, which is used for `stream.mark`.
2. The `entryCnt` is used for storing the number of entries we have read.
3. I modified `parseEntry` a little bit : nothing would be written to `xhtml` if a zip entry uses `STORED` and `Data Descriptor` at the same time. Instread an exception is thrown and the stream would be `reset` and read for a second time.
4. I have generated a zip archive for test. This zip contains 5 entries. The 2nd and 4th entry in the zip archive are using  `STORED` with `Data Descriptor`. This zip archive could be successfully parsed.

See also [#356](https://github.com/apache/tika/pull/356) and [Commons Compress #137](https://github.com/apache/commons-compress/pull/137) for more information.


tbentleypfpt:
Awesome! Could this be released in a 1.24.2 or 1.25 release soon?

tballison:
@PeterAlfredLee, thank you for this PR.  The one small item that I had to fix was using an instance variable in PKGParser.  Parsers have to be thread-safe.  Otherwise, tho, this was an elegant solution.  Thank you.

PeterAlfredLee:
I'm happy this helps a little bit. :)

lfcnassif:
This have broken thread safety of PackedParser because of the created instance field, it should be passed through method params.

tballison:
Ugh...sorry about that!

PR #366, opened 2020-09-30
Fix build fail caused by can't find test file

PeterAlfredLee:
Test with [latest commit 75c2ff5](https://github.com/apache/tika/commit/75c2ff5686a70c0fb15c4b52534c1be09669af1b) in `main` branch and got some test fail caused by can't find test file.
This PR is a fix for it.


Gagravarr:
Thanks for spotting this, the perils of cherry-picking between 1.x and 2.x branches without paying enough attention!

I think this might have already been fixed now, can you please confirm if we still need any parts of your fix?

PeterAlfredLee:
Everything looks good to me. Thank you for reviewing this.

PR #375, opened 2020-11-02
Add CompareUtils and simplify some sort

PeterAlfredLee:
Add CompareUtils, remove duplicate code and simplify some sort.


kkrugler (approved):
lgtm, nice cleanup

PR #376, opened 2020-11-03
Simplify some sort statement use List.sort and lambda

PeterAlfredLee:


kkrugler (approved):
Assuming all tests still pass, this seems like reasonable cleanup of the code.

PR #377, opened 2020-11-07
Simplify some equals statement by use Objects.equals

PeterAlfredLee:


tika-eval/src/main/java/org/apache/tika/eval/EvalFilePaths.java
```diff
@@ -81,9 +82,8 @@ public boolean equals(Object o) {
 
         if (sourceFileLength != that.sourceFileLength) return false;
         if (extractFileLength != that.extractFileLength) return false;
-        if (relativeSourceFilePath != null ? !relativeSourceFilePath.equals(that.relativeSourceFilePath) : that.relativeSourceFilePath != null)
-            return false;
-        return extractFile != null ? extractFile.equals(that.extractFile) : that.extractFile == null;
+        if (!Objects.equals(relativeSourceFilePath, that.relativeSourceFilePath)) return false;
```
kkrugler, line 85: In general I think Tika uses a "always braces" style, so this would be better as:

```java
if (!blah) {
    return false;
} else {
    return Objects.equals(blah);
}
```

tika-eval/src/main/java/org/apache/tika/eval/db/ColInfo.java
```diff
@@ -100,8 +101,8 @@ public boolean equals(Object o) {
 
         if (type != colInfo.type) return false;
         if (name != colInfo.name) return false;
-        if (precision != null ? !precision.equals(colInfo.precision) : colInfo.precision != null) return false;
-        return !(constraints != null ? !constraints.equals(colInfo.constraints) : colInfo.constraints != null);
+        if (!Objects.equals(precision, colInfo.precision)) return false;
```
kkrugler, line 104: Same comment as above

tika-parser-modules/tika-parser-pdf-module/src/main/java/org/apache/tika/parser/pdf/PDFMarkedContent2XHTML.java
```diff
@@ -500,8 +500,8 @@ public boolean equals(Object o) {
 
             HtmlTag htmlTag = (HtmlTag) o;
 
-            if (tag != null ? !tag.equals(htmlTag.tag) : htmlTag.tag != null) return false;
-            return clazz != null ? clazz.equals(htmlTag.clazz) : htmlTag.clazz == null;
+            if (!Objects.equals(tag, htmlTag.tag)) return false;
+            return Objects.equals(clazz, htmlTag.clazz);
```
kkrugler, line 504: Same comment as above

tika-eval/src/main/java/org/apache/tika/eval/EvalFilePaths.java
PeterAlfredLee: Sure. And this is also the style I prefer. I made the `if` clause like this cause I want to have the same style as line 83, 84. Will modify all these 3 lines.

kkrugler (approved):
lgtm

PR #378, opened 2020-11-08
fix for TIKA-3225 contributed by onozaty

onozaty:
Added a namespace declaration because it may cause an error when processing XML because there is no namespace declaration.


tballison:
LGTM. Any fellow devs object?

Gagravarr:
The mime-info xml format was/is a standard[1], but we long ago added enough of our own matching types that other programs won't be able to cope fully with our file so we probably don't need to worry about externally testing the impact. Given the proposed change is standard xml, looks fine to me

[1] https://specifications.freedesktop.org/shared-mime-info-spec/latest/ar01s02.html#idm45164881929088

PR #379, opened 2020-11-12
Fix unecessary close

lcc:
Meaningless Close: In several classes, close(), specified in Closeable interface, has no effect and other methods in those classes can be called after close() without IOException.

### DIFF
--- a/tika-core/src/test/java/org/apache/tika/config/ParamTest.java
+++ b/tika-core/src/test/java/org/apache/tika/config/ParamTest.java
@@ -61,8 +61,6 @@ public class ParamTest {
             ByteArrayOutputStream stream = new ByteArrayOutputStream();
             param.save(stream);
             ByteArrayInputStream inStream = new ByteArrayInputStream(stream.toByteArray());
-            stream.close();
-            inStream.close();
             Param<?> loaded = Param.load(inStream);
             assertEquals(param.getName(), loaded.getName());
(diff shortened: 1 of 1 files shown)


tika-core/src/test/java/org/apache/tika/config/ParamTest.java
```diff
@@ -61,8 +61,6 @@ public void testSaveAndLoad() throws Exception {
             ByteArrayOutputStream stream = new ByteArrayOutputStream();
             param.save(stream);
             ByteArrayInputStream inStream = new ByteArrayInputStream(stream.toByteArray());
-            stream.close();
-            inStream.close();
```
kkrugler, line 65: The call to `inStream.close()` is confusing here, as it's used in the next line. You're right that a `close()` for either stream does nothing, but to avoid false positive warnings from tools that aren't so smart, I would just move these two calls to the end of the loop.

lcc: Sorry for the late response, can you tell me which tools send a false positive warning? Also, would you like me to put these two calls at the end of the loop?

kkrugler: Findbugs would (or at least used to) always report a missing close on any kind of InputStream (anything that was closable). And yes, I was suggesting putting those two calls at the end of the loop.

PR #380, opened 2020-11-12
Add two tests for `OPCPackageDetector`

PeterAlfredLee:
Add two test for method `OPCPackageDetector#detect`


tballison:
Do these duplicate any existing tests?

PeterAlfredLee:
> Do these duplicate any existing tests?

No. But these classes are covered by other tests already. The tests here are just directly calling OPCPackageDetector to have a test.

PR #382, opened 2020-11-12
Simplify some code in OPCPackageDetector#detect

PeterAlfredLee:
1. Simpify catch statement in method `detect`.
2. Adjust some comment in method `detect`.


tika-parser-modules/tika-parser-microsoft-module/src/main/java/org/apache/tika/detect/microsoft/ooxml/OPCPackageDetector.java
```diff
@@ -159,32 +159,32 @@
...
-            closeQuietly(zipEntrySource);
-            closeQuietly(zipFile);
-            //TIKA-2571
-            throw e;
         } catch (InvalidFormatException | RuntimeException e) {
             closeQuietly(zipEntrySource);
             closeQuietly(zipFile);
-            return null;
+
+            if (e instanceof SecurityException) {
```
tballison, line 180: I have a personal preference for catching/handling specific exceptions according to behavior of the response.  Is there a performance reason to check `instanceof` SecurityException?

PeterAlfredLee: Ah, I was just thinking to simplify the two catch clause as they look similar. Haven't think about the performance here. :)

PR #383, opened 2020-11-14
Simplify some call of String operation

PeterAlfredLee:
1. Simplify some call of `StringBuilder#append`.
2. Simplify some call of `String#substring`.
3. Remove unnecessary call of `#toString`.


epugh:
Nice cleanups!!

PR #385, opened 2020-11-17
Simplify some test code by using method TikaTest#getResourceAsUrl and TikaTest#getResourceAsUri

PeterAlfredLee:
1. Add method `getResourceAsUrl` to `TikaTest`
2. Add method `getResourceAsUri` to `TikaTest`
3. Simplify some test code by using method `TikaTest#getResourceAsUrl` and `TikaTest#getResourceAsUri`

### DIFF
--- a/tika-core/src/test/java/org/apache/tika/TikaTest.java
+++ b/tika-core/src/test/java/org/apache/tika/TikaTest.java
@@ -62,6 +62,15 @@ public abstract class TikaTest {
 
     protected static Parser AUTO_DETECT_PARSER = new AutoDetectParser();
 
+    /**
+     * Finds a resource with a given name.
+     * @param name name of the desired resource
+     * @return A {@link java.net.URL} object or null
+     */
+    public URL getResourceAsUrl(String name) {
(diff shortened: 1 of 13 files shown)


tika-core/src/test/java/org/apache/tika/TikaTest.java
```diff
@@ -62,6 +62,15 @@
 
     protected static Parser AUTO_DETECT_PARSER = new AutoDetectParser();
 
+    /**
+     * Finds a resource with a given name.
+     * @param name name of the desired resource
+     * @return A {@link java.net.URL} object or null
+     */
+    public URL getResource(String name) {
```
kkrugler, line 70: A more descriptive name - like `getResourceAsUrl()` - would help when reading the code. Wondering if a `getResourceAsUri()` should also be added.

PeterAlfredLee: Sure. Will update change.

kkrugler (approved):
lgtm!

PR #387, opened 2020-11-23
Simplify method Tika#parseToString(InputStream, Metadata)

PeterAlfredLee:
Replace with `return parseToString(stream, metadata, maxStringLength);`


kkrugler (approved):
lgtm

PR #392, opened 2020-12-12
Simplify method CompressorParser#parse

PeterAlfredLee:


THausherr (approved):

tika-parsers/tika-parsers-classic/tika-parsers-classic-modules/tika-parser-pkg-module/src/main/java/org/apache/tika/parser/pkg/CompressorParser.java
```diff
@@ -213,22 +213,15 @@ public boolean decompressConcatenated(Metadata metadata) {
...
-                    name = name.substring(0, name.length() - 4);
-                } else if (name.endsWith(".xz")) {
-                    name = name.substring(0, name.length() - 3);
-                } else if (name.endsWith(".zlib")) {
-                    name = name.substring(0, name.length() - 5);
-                } else if (name.endsWith(".pack")) {
-                    name = name.substring(0, name.length() - 5);
-                } else if (name.endsWith(".br")) {
-                    name = name.substring(0, name.length() - 3);
+                if (name.endsWith(".tbz") || name.endsWith(".tbz2")) {
```
ADarko22, line 216: We could have a static method
` 
  public static boolean endsWithAnyOf(String string, String... suffixes) {
    return Stream.of(suffixes)
        .filter(Objects::nonNull)
        .anyMatch(string::endsWith);
  }
`
and  then
`
if(endsWithAnyOf(name, ".tbz", ".tbz2")){
..
}else if(endsWithAnyOf(name, ".bz", ".bz2", ".xz", ".zlib", ".pack", ".br")){
..
}
`
not sure where could be the best place for: `public static boolean endsWithAnyOf(String string, String... suffixes)`

PeterAlfredLee: Well, I don't think this method would be reused in anyplace else.

PR #400, opened 2021-01-27
TIKA-3282 fix non-ascii characters in onenote

AdrianD-intrafind:
I changed the characterset and added a unit test for this issue.


nddipiazza:
looks good to me @tballison

tballison:
Thank you @AdrianD-intrafind!

PR #402, opened 2021-02-03
Tika 3286 - Check if lanaguage files exists and provide better error msg; Support script lanaguge files in script directory

peterkronenberg:


tika-parsers/tika-parsers-classic/tika-parsers-classic-modules/tika-parser-ocr-module/src/main/java/org/apache/tika/parser/ocr/TesseractOCRConfig.java
```diff
@@ -249,17 +255,71 @@ public String getLanguage() {
...
+            throw new IllegalArgumentException("Invalid language code(s): " + invalidCodes);
         }
         this.language = language;
     }
 
+
+    /**
+     * Check if tessdata language model exists
+     */
+    private boolean langExists(String lang) {
```
tballison, line 295: If the user doesn't specify a lang dir, I don't think we should make assumptions or go looking for it.

On my ubuntu laptop, tessdata is under /usr/share/tesseract-ocr/4.00/tessdata _not_ linux default is.  On my mac, tessdata isn't under /usr/share/tessdata, and as you've pointed out, on my Windows laptop, who knows where tessdata is.

peterkronenberg: And you're not specifying the path?  I guess you're right.  The user can put it anywhere and just make sure the Path points to it.  So perhaps I should search the path for both xNix and Windows?  For Windows, I search for Tesseract and assume tessdata is under it.  For Unix, I just search for tessdata on the path?

tballison: I don't think this feature offers enough to justify the code (admittedly small) and the maintenance.  I'm extremely grateful for the updates to the script checking, and I'm happy that we now throw an exception that includes stderr if tesseract has an exitValue of > 0.

peterkronenberg: So you're saying that if the language directory is not specified by the user, for any OS, then we just pass it to Tesseract and let Tesseract handle it?  (i.e., what I'm currently doing for Windows)

tballison: Yes.

peterkronenberg: done

tballison:
Thank you!

PR #403, opened 2021-02-08
Allow tesseract/tessdata path to be specified by environment variables

peterkronenberg:


lfcnassif:
I think this logic should be inverted to use ENV var if property or tesseractConfig object is not set in context. So global system configuration could be overridden by app specific configuration.

peterkronenberg:
Hi, not sure I follow.   Are you saying TesseractConfig should override?  If the environment variable isn't set, then it will have the same behavior as before.  If you specify the env variables, than they will override

lfcnassif:
Yes, inverting the logic, if tesseractPath (or tessdataPath) has the default empty value, env var would be used if set. If properties are specified, it will keep past behaviour too. I'm thinking about different apps in the same system using Tika, but with different configurations. Inverting the logic, it will not be an issue.

peterkronenberg:
I think that's what it's already doing.  If you don't specify it in the config and it is specified in the environment, it will use the environment values.

peterkronenberg:
So I think what you're saying is that even if environment variables are specified, then still use the config by default.  And only if config is blank, then use environment ?

lfcnassif:
> So I think what you're saying is that even if environment variables are specified, then still use the config by default. And only if config is blank, then use environment ?

Yes that was my suggestion. Other commiter can have other opinion of course.

peterkronenberg:
I just want to make sure that it doesn't break anything or have any unexpected behavior.  Let me try it your way and see how it works

peterkronenberg:
Overtaken by events

PR #410, opened 2021-03-04
TIKA-3310 Check if MP4 file's compatible brands match any of the expected values

peterkronenberg:
…cted values


peterkronenberg:
This is ready to merge

PR #412, opened 2021-03-13
TIKA-3317 - add a solr fetch iterator to tika pipes

nddipiazza:
Tika pipes needs a solr fetch iterator. Some of the parameters I have in this may no longer be needed (or will be different) given how tika pipes is different from my tika async code i wrote for Tika 1.x. We can change it as we need. 

Change the solr emitter to use solrj. 

Add integration test that uses TestContianers + docker so we can do real test against solr6 7 and 8 

TODO: 

- [ ] add atomic update options, and respect each of the options listed here: https://solr.apache.org/guide/6_6/updating-parts-of-documents.html#UpdatingPartsofDocuments-OptimisticConcurrency
- [ ] add a unit test for each different update option for solr emitter


nddipiazza:
hey @tballison  i merged main into the branch. I still want to spend some time with Update options. But need to ask some questions to my solr guru friends to bless.

nddipiazza:
closed in lieu of https://github.com/apache/tika/pull/445

PR #415, opened 2021-03-16
Minor cleaning and added missing javadoc on TikaServer

Subhajitdas298:
Minor restructure and added missing javadoc on TikaServer.
No functional change.

Modified existing methods based on usage for clearer code.


tballison:
Thank you @Subhajitdas298 !

PR #419, opened 2021-03-17
fix for TIKA-3329 contributed by Thamme Gowda

thammegowda:
RTG translator (https://isi-nlp.github.io/rtg/) integration to tika-translate 

This service requires running RTG model for 500-eng translation as a REST service

By default, the RTGTranslator uses `http://localhost:6060` as base URL 

Which can be obtained from a prebuilt docker image:

* docker image can be run on CPU (i.e. without GPU, for testing):
    `docker run --rm -i -p 6060:6060 tgowda/rtg-model:500toEng-v1`
   
* but GPU (e.g. Device 0) is recommended for translating a lot of documents:
    `docker run --rm -i -p 6060:6060 --gpus '"device=0"' tgowda/rtg-model:500toEng-v1`
    
`rtg.base.url` property from `translator.rtg.properties` file can be  used to customize base URL if needed. 

CC @chrismattmann  @tballison

### DIFF
--- a/tika-translate/src/main/java/org/apache/tika/language/translate/RTGTranslator.java
+++ b/tika-translate/src/main/java/org/apache/tika/language/translate/RTGTranslator.java
@@ -81,7 +81,7 @@ public class RTGTranslator extends AbstractTranslator {
     public RTGTranslator() {
         String rtgBaseUrl = RTG_TRANSLATE_URL_BASE;
         Properties config = new Properties();
-        try (InputStream stream = getClass().getResourceAsStream(RTG_PROPS)){
+        try (InputStream stream = getClass().getClassLoader().getResourceAsStream(RTG_PROPS)){
             if (stream != null){
                 config.load(stream);
             }
@@ -92,8 +92,13 @@ public class RTGTranslator extends AbstractTranslator {
         LOG.info("RTG base URL: " + rtgBaseUrl);
         List<Object> providers = new ArrayList<>();
         providers.add(new JacksonJsonProvider());
-        this.client = WebClient.create(rtgBaseUrl, providers);
-        this.isAvailable = client.head().getStatus() == 200;
+        try {
+            this.client = WebClient.create(rtgBaseUrl, providers);
+            this.isAvailable = client.head().getStatus() == 200;
+        } catch (Exception e){
+            LOG.warn(e.getMessage(), e);
+            isAvailable = false;
+        }
 
     }
     @Override
--- a/tika-translate/src/main/java/org/apache/tika/language/translate/RTGTranslator.java
+++ b/tika-translate/src/main/java/org/apache/tika/language/translate/RTGTranslator.java
@@ -1,0 +1,137 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package org.apache.tika.language.translate;
+
+import com.fasterxml.jackson.jaxrs.json.JacksonJsonProvider;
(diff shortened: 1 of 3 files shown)


lewismc:
@thammegowda some accompanying documentation on the wiki would be VERY useful for this.

thammegowda:
@lewismc Thanks for the suggestion. I will make a page on the wiki for this feature.

thammegowda:
Wiki page created:  https://cwiki.apache.org/confluence/display/TIKA/NMT-RTG

tika-translate/src/main/java/org/apache/tika/language/translate/RTGTranslator.java
```diff
@@ -0,0 +1,137 @@
...
+    public static final String RTG_TRANSLATE_URL_BASE = "http://localhost:6060";
+    public static final String RTG_PROPS = "translator.rtg.properties";
+    private static final Logger LOG = LoggerFactory.getLogger(RTGTranslator.class);
+    private WebClient client;
+    private boolean isAvailable = false;
+
+    public RTGTranslator() {
+        String rtgBaseUrl = RTG_TRANSLATE_URL_BASE;
+        Properties config = new Properties();
+        try (InputStream stream = getClass().getResourceAsStream(RTG_PROPS)){
```
kieraCurtis: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [UI: Usage of GetResource may be unsafe if class is extended (UI_INHERITANCE_UNSAFE_GETRESOURCE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#ui-usage-of-getresource-may-be-unsafe-if-class-is-extended-ui-inheritance-unsafe-getresource).
Calling this.getClass().getResource(...) could give results other than expected if this class is extended by a class in another package.

thammegowda: @kieraCurtis Thanks for the comment. Agreed, this was a problematic statement, and hence I have already changed it to `getClass().getClassLoader()` in https://github.com/apache/tika/pull/419/commits/9068145e48e53f5e355671abdce47d993e7f790c . 
The behaviour of `getClass().getClassLoader().getResource*` is going to be same regardless of package of class. The only exceptional case is if classloaders are swapped between parents and children classes e.g. plugin systems that isolate classpaths for each plugin. Since Tika system is not messing with its classloaders; all classes have the same classloader, and hence my revised statement is NOT BAD_PRACTICE. 

That being said, I do not know better, safer, more robust alternatives; if there are suggestions, I am happy to revise my code :) 
 

P.S. 
1.  Last time I wrote java code was 4+ years ago, this thing had slipped in by mistake and I immediately corrected it. 
1. The code you are reviewing/commenting is outdated.

kieraCurtis: thank you for your quick reply. Happy

lewismc:
+1 @thammegowda excellent job

chrismattmann:
going to test this today in Tika. If everything passes, I'll get it committed and then work to integrate it directly into tika python as the default translation package. Thanks @thammegowda !

chrismattmann:
OK had to make some changes so that it would pass the forbidden APIs (RTG and RTGTest) and also to tika-server classic modules that were failing checkstyle (probably has for a while but my Maven version seems to care and have checkstyle as a failure case which I haven't seen). Also had to update the test b/c the translation returned slightly different than the original PR (it had an extra comma, and a period). Anyways it's fixed and works!

```
[INFO] tika-server-classic ................................ SUCCESS [ 16.967 s]
[INFO] tika-server-client ................................. SUCCESS [  1.039 s]
[INFO] Apache Tika eval ................................... SUCCESS [  0.063 s]
[INFO] tika-eval-core ..................................... SUCCESS [ 13.070 s]
[INFO] tika-eval-app ...................................... SUCCESS [ 21.834 s]
[INFO] Apache Tika fuzzing ................................ SUCCESS [  0.938 s]
[INFO] Apache Tika examples ............................... SUCCESS [  9.334 s]
[INFO] Apache Tika Java-7 Components ...................... SUCCESS [  1.628 s]
[INFO] Apache Tika ........................................ SUCCESS [  0.024 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  10:42 min
[INFO] Finished at: 2021-05-01T13:41:08-07:00
[INFO] ------------------------------------------------------------------------
[2]-  Done                    emacs tika-translate/src/test/java/org/apache/tika/language/translate/RTGTranslatorTest.java
[3]+  Done                    emacs RTGTranslator.java  (wd: ~/git/tika/tika-translate/src/main/java/org/apache/tika/language/translate)
(wd now: ~/git/tika)
(base) mattmann@proscuitto:~/git/tika$ 
```

PR #430, opened 2021-04-22
[TIKA-3357] Remove ambiguity in request handlers - main

Subhajitdas298:
Added Resource comparator based to produce type.
In an ambiguous call, request handler will be chosen based on the type of data it returns.

*Current priority is set as:*
MediaType.TEXT_PLAIN_TYPE,
MediaType.APPLICATION_JSON_TYPE,
MediaType.TEXT_HTML_TYPE,
MediaType.TEXT_XML_TYPE
*The lower in list (higher index value), the higher priority it has.*
In case of no matching in this list, it will be treated as media type all.

*Note: Please change priority list if some other order is more suitable.*

This is kept at sync with branch_1x


lewismc:
@Subhajitdas298 I didn't get a chance to try yet. Does this PR address the annoying logging output we see when starting tika-server?

Subhajitdas298:
> @Subhajitdas298 I didn't get a chance to try yet. Does this PR address the annoying logging output we see when starting tika-server?

Hi, this does not affect any startup logs. This only deals with, choice of REST handler method. When multiple handlers are valid, say in case of "Accept: */*" header, this priority list will decide which one to choose.

PR #432, opened 2021-04-24
[TIKA-3368] Add Bill of Materials (BOM) artifact (Tika 1.x)

grossws:
Fixes #TIKA-3368


lewismc (approved):
+1 I like BOM and after clearing `~/.m2` cache I was able to build and test 1.x locally with this.
I'm not sure about the XML formatting. I thought it was 4-space indents but this is not a blocker.

grossws:
Since Tika 1.x will reach EOL in September, 2022 I think this PR is no longer relevant.

PR #433, opened 2021-04-28
[TIKA-3374] Apply charset detection for archive entry name

Ryan421:
Fixes #TIKA-3374


tika-parsers/tika-parsers-classic/tika-parsers-classic-modules/tika-parser-pkg-module/src/main/java/org/apache/tika/parser/pkg/PackageParser.java
```diff
@@ -392,6 +392,15 @@ private void parseEntry(ArchiveInputStream archive, ArchiveEntry entry,
                             XHTMLContentHandler xhtml)
             throws SAXException, IOException, TikaException {
         String name = entry.getName();
+        
+        //Try to detect charset of archive entry in case of non-unicode filename is used
+        if (entry instanceof ZipArchiveEntry) {
+            detector.setText(((ZipArchiveEntry) entry).getRawName());
```
tballison, line 398: This is embarrassing, but where is the detector initialized...how are you getting it?

Maybe have PackageParser extend AbstractEncodingDetectorParser?

Ryan421: Yes, it is really embarrasssing, will change to extend AbstractEncodingDetectorParser and using getEncodingDetector to do the job, thank you so much.

Ryan421:
Add unit test with a dummy EncodingDetector to verify the charset detection flow is executed.

tika-parsers/tika-parsers-classic/tika-parsers-classic-modules/tika-parser-pkg-module/src/main/java/org/apache/tika/parser/pkg/PackageParser.java
tballison: Sorry, please forgive me.  I meant embarrassing for me because I figured I was missing something!!!

Ryan421: No need to be sorry ^^, It was really my fault when moving the code block from our project to here and not properly checked. Really appreciate your review and suggestions.

PR #436, opened 2021-05-08
fix for TIKA-3389 contributed by kamaci

kamaci:


kamaci:
Some resources should not be closed and I've didn't close them. This PR passes all the tests right now. Anyone is welcome to review for such cases.

PR #437, opened 2021-05-11
fix for TIKA-3390 contributed by kamaci

kamaci:


tika-parsers/tika-parsers-classic/tika-parsers-classic-modules/tika-parser-microsoft-module/src/main/java/org/apache/tika/parser/microsoft/onenote/ExtendedGUID.java
```diff
@@ -35,7 +35,7 @@ public static ExtendedGUID nil() {
     @Override
     public int compareTo(ExtendedGUID other) {
         if (other.guid.equals(guid)) {
-            new Long(n).compareTo(other.n);
+            return Long.compare(n, other.n);
```
kamaci, line 38: I've changed it like that since it seems like a bug.

PR #441, opened 2021-05-12
fix for TIKA-3400 contributed by kamaci

kamaci:


tika-eval/tika-eval-core/src/main/java/org/apache/tika/eval/core/tokens/URLEmailNormalizingFilterFactory.java
```diff
@@ -69,11 +69,10 @@ public boolean incrementToken() throws IOException {
                 return false;
             }
             //== is actually substantially faster than .equals(String)
-            if (typeAtt.type() == UAX29URLEmailTokenizer.TOKEN_TYPES[UAX29URLEmailTokenizer.URL]) {
+            if (typeAtt.type().equals(UAX29URLEmailTokenizer.TOKEN_TYPES[UAX29URLEmailTokenizer.URL])) {
```
tballison, line 72: This was done out of a notional sense of efficiency. I'm not sure we need to change it.

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-microsoft-module/src/main/java/org/apache/tika/parser/microsoft/onenote/FileNode.java
```diff
@@ -257,11 +257,11 @@ public void print(OneNoteDocument document, OneNotePtr pointer, int indentLevel)
                     subType.revisionManifest.revisionRole);
 
         }
-        if ((gctxid != ExtendedGUID.nil() ||
+        if ((!gctxid.equals(ExtendedGUID.nil()) ||
```
tballison, line 260: Good catch!  We should probably make a static constant ExtendedGUID.NIL to avoid unnecessary object creation.

tika-eval/tika-eval-core/src/main/java/org/apache/tika/eval/core/tokens/URLEmailNormalizingFilterFactory.java
kamaci: So, parameter of the `TypeAttribute#setType` can be exactly that String (`UAX29URLEmailTokenizer.TOKEN_TYPES[UAX29URLEmailTokenizer.URL]`) ?

tballison: This relies on the Lucene not changing the underlying static strings: https://github.com/apache/lucene/blob/main/lucene/analysis/common/src/java/org/apache/lucene/analysis/email/UAX29URLEmailTokenizer.java#L61

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-microsoft-module/src/main/java/org/apache/tika/parser/microsoft/onenote/FileNode.java
tballison: To be clear, I'm not asking you to do the static thing on this issue.  Your catch is important.  Thank you!

tika-eval/tika-eval-core/src/main/java/org/apache/tika/eval/core/tokens/URLEmailNormalizingFilterFactory.java
kamaci: OK. I think that they had to use enum instead of a string array for such a thing :blush: I'll rollback that lines at my PR.

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-microsoft-module/src/main/java/org/apache/tika/parser/microsoft/onenote/FileNode.java
kamaci: You are welcome!

kamaci:
@tballison I've updated the PR. Checks fail due to `MP4ParserTest.java:101`. I don't get same exception at my local environment. Do you have any idea about the reason?

tballison:
I was just able to replicate that in Java 11 on a Mac.  ubuntu w Java 8 passes... Ugh... I pushed a simple fix for now.

kamaci:
@tballison is there anything left to do for this PR?

PR #467, opened 2021-12-20
[TIKA-3627] Fix OOXML parsing for multiple threads

bgeisberger:
Fixes TIKA-3627


tballison:
This is critical.  I'm going to respin 2.2.1.  Thank you for finding this!

PR #471, opened 2022-01-01
initial results of static analysis

dk2k:
Analysis was performed with Findbugs


tballison:
Thank you for opening this.  I don't want to add more work for you, but this PR is too big for me to handle in one go.

Can we break it up in PRs either by module or by severity?  Or better yet, add a findbugs maven plugin to our project so that this isn't a fix-once-and-hope-for-the-best-in-the-future kind of thing?

dk2k:
I expect some opposition from other members of the team: Findbugs isn't maintained any longer and maven reports a warning about the structure of pom.xml when I add a section for findbugs. As for separate change lists for modules, I think it's feasible.

tballison:
I used to use Findbugs on other projects and admire many of its features.  After you opened this and I made my comment, I went digging a bit and found what you point out: no longer supported.  I'd like to make many of the changes you found, but I'd also like to make this more of a lasting contribution somehow.  Maybe sonarqube, e.g. https://sonarcloud.io/project/overview?id=apache_any23 ???

dk2k:
I will check sonarqube
Meanwhile, PR for module CORE https://github.com/apache/tika/pull/478

PR #486, opened 2022-01-22
TIKA-3649 fixes for static analysis report of module tika-translate

dk2k:
Usage of StringBuffer in one thread method was an overkill


lewismc:
Thank you @dk2k

PR #520, opened 2022-03-03
Fix email detection (TIKA-3687)

SchwingSK:
1024  is maybe a bit overkill for the X|DKIM|ARC headers lookahead ?

### DIFF
--- a/tika-core/src/main/resources/org/apache/tika/mime/tika-mimetypes.xml
+++ b/tika-core/src/main/resources/org/apache/tika/mime/tika-mimetypes.xml
@@ -6418,6 +6418,9 @@
         <match value="\nUser-Agent:" type="string" offset="0:1024"/>
         <match value="\nX-Mailer:" type="string" offset="0:1024"/>
         <match value="\nX-Originating-IP:" type="stringignorecase" offset="0:1024"/>
+        <match value="\nX-" type="string" offset="0:1024"/>
+        <match value="\nDKIM-" type="string" offset="0:1024"/>
+        <match value="\nARC-" type="string" offset="0:1024"/>        
       </match>
       <!-- match X- DKIM- ARC- at start of file and then require at least one
            of the usual: from, received, date...but look farther into the file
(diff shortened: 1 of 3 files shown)


tika-core/src/main/resources/org/apache/tika/mime/tika-mimetypes.xml
```diff
@@ -6422,7 +6422,7 @@
       <!-- match X- DKIM- ARC- at start of file and then require at least one
            of the usual: from, received, date...but look farther into the file
            because of the X|DKIM|ARC headers-->
-      <match value="(X|DKIM|ARC)-" type="regex" offset="0">
+      <match value="(X|DKIM|ARC)-" type="regex" offset="0:1024">
```
tballison, line 6425: I worry about looking for X- anywhere in the first 1024 without requiring a \n before it.

What would you think of adding something like this into the previous minShouldMatch=2 clause?
`
<match value="\nX-" type="string" offset="0:1024">
<match value="\nDKIM-" type="string" offset="0:1024">
<match value="\nARC-" type="string" offset="0:1024">
`

SchwingSK: Good point, I like your solution better. Code changed accordingly.

lfcnassif:
Sorry, late to the party... I think maybe the new \nX- 0:1024 could cause false positives in regression tests. If yes, I would check for additional headers after it matches, like done for mbox mimetype.

lfcnassif:
Also, if we look for additional headers if X|DKIM|ARC matches at the beginning, why not look for other headers if they match in a wide range of positions? (ok there is the \n but it is pretty common in txt files).

Maybe we could use a regex to put all definitions together, regardless if they are at the beginning at the file or at the beginning of lines in the first 1024...

lfcnassif:
I just saw the minShouldMatch=2 clause (I was on the small cellphone screen...), let's wait for the regression tests...

lfcnassif:
@SchwingSK I detected some false positives (see TIKA-3687) after this. The <match value="\nX-" type="string" offset="0:1024"/> rule is not needed to detect the EML sample you provided in TIKA-3687. Not sure if this is the best option, but if we remove it, will EML files in your dataset continue to be detected?

PR #552, opened 2022-04-25
TIKA-1800 -- decode the escape character before parsing it as a new p…

longphan98:
…arameter

* the issue ID (`TIKA-XXXX`)


longphan98:
And also a test case too :3

Gagravarr:
Your commit seems to remove a test zip file, which seems to be by accident, any chance you could re-do it without the accidental deletion?

Would you also be able to add a comment explaining what your new for loop is doing, so anyone looking at that code later can quickly figure out what it's doing?

PR #554, opened 2022-05-03
TIKA-XXXX returns maximum mime priority in case of collision

vrnvu:
If two mime types collision, we will always return the first element in the queue.

https://github.com/vrnvu/tika/blob/36b0379ffb73b6712dd58156b87f1df5d556b019/tika-core/src/main/java/org/apache/tika/mime/MimeTypes.java#L580

In order to make use of the priority (and have an actual "priority queue") we need to always insert at the index 0. This way the highest priority gets returned.

TODO:

* the issue ID (`TIKA-XXXX`)

### DIFF
--- a/tika-core/src/main/java/org/apache/tika/mime/MimeTypes.java
+++ b/tika-core/src/main/java/org/apache/tika/mime/MimeTypes.java
@@ -234,7 +234,7 @@ public final class MimeTypes implements Detector, Serializable {
                 break;
             }
             if (magic.eval(data)) {
-                result.add(magic.getType());
+                result.add(0, magic.getType());
                 currentPriority = magic.getPriority();
             }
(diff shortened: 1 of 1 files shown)


nick-burch-flec:
Are you able to put together a very small junit unit test showing the problem you're facing and the fix with your patch applied? That will help us ensure it stays fixed in future!

vrnvu:
Sure, we encountered the error in production with an image that had this collision. I'll take a look at what unit test I need to write and get back to you. I created the PR just to see what we need to do to contribute with the fix. ^^

vrnvu:
I'm closing the PR. Seems that in newer versions of TIKA than the one we were using, the list of priorities is sorted. So this fix is not needed.

PR #558, opened 2022-05-12
Initial commit for review.

monkmachine:

* the issue ID (`TIKA-XXXX`)


tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-cad-module/src/main/java/org/apache/tika/parser/dwg/DWGReadParser.java
```diff
@@ -0,0 +1,208 @@
...
+                            parseHeader(jParser,metadata);
+                        } else {
+                            jParser.skipChildren();
+                        }
+                    }
+                }
+            }
+            jParser.close();
+        } finally {
+            FileUtils.deleteQuietly(tmpFileOut);
```
tballison, line 126: Should also delete tmpfileoutcleaned in case there's an exception?

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-cad-module/src/main/java/org/apache/tika/parser/dwg/DWGReadParser.java
```diff
@@ -0,0 +1,208 @@
...
+        xhtml.startDocument();
+        UUID uuid = UUID.randomUUID();
+        File tmpFileOut = File.createTempFile(uuid + "dwgreadout", ".json");
+        File tmpFileOutCleaned = File.createTempFile(uuid + "dwgreadoutclean", ".json");
+        File tmpFileIn = File.createTempFile(uuid + "dwgreadin", ".dwg");
+        try {
+
+            FileUtils.copyInputStreamToFile(stream, tmpFileIn);
+
+            List < String > command = Arrays.asList(dwgc.getDwgReadExecutable(), "-O", "JSON", "-o",
```
tballison, line 63: Consider using https://github.com/apache/tika/blob/main/tika-core/src/main/java/org/apache/tika/utils/ProcessUtils.java#L165 instead of hand rolling ProcessBuilder?

Definitely need a timeout and should never waitFor() forever.

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-cad-module/src/main/java/org/apache/tika/parser/dwg/DWGReadParser.java
```diff
@@ -0,0 +1,208 @@
...
+            Process p = new ProcessBuilder(command).start();
+
+            try {
+                int exitCode = p.waitFor();
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+            }
+
+            if (dwgc.isCleanDwgReadOutput()) {
+                // dwgread sometimes creates strings with invalid utf-8 sequences or invalid json (nan instead of NaN). replace them
```
tballison, line 74: Oh, my head hurts.  Thank you for this.

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-cad-module/src/main/java/org/apache/tika/parser/dwg/AbstractDWGParser.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+    String getCleanDwgReadReplaceWith() {
+        return defaultDwgParserConfig.getCleanDwgReadReplaceWith();
+    }
+    
+    @Field
+    public void setCleanDwgReadReplaceWith(String cleanDwgReadReplaceWith) {
+    	defaultDwgParserConfig.setCleanDwgReadReplaceWith(cleanDwgReadReplaceWith);
+    }
+
```
tballison, line 84: May want to implement Initializable and see if the dwg executable is actually working.  Please add lots of logging if a user specified a dwg executable, but it isn't working.

Something along the lines of: https://github.com/apache/tika/blob/main/tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-ocr-module/src/main/java/org/apache/tika/parser/ocr/TesseractOCRParser.java#L529

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-cad-module/src/test/java/org/apache/tika/parser/dwg/DWGParserTest.java
```diff
@@ -29,166 +29,179 @@
...
-        Metadata metadata = getXML("testDWG-AC1027.dwg").metadata;
-        assertEquals("hlu", metadata.get(TikaCoreProperties.MODIFIER));
-    }
-
-    @Test
-    public void testAC1032() throws Exception {
-        Metadata metadata = getXML("testDWG-AC1032.dwg").metadata;
-        assertEquals("jlakshvi", metadata.get(TikaCoreProperties.MODIFIER));
-        assertEquals("CUSTOMER'S ADDRESS", metadata.get("dwg-custom:CUSTOMER'S ADDRESS"));
+    public void testDWGRead() throws Exception {
```
tballison, line 198: If the executable is required, we should "assume" that it exists at the start of the test. Along the lines: https://github.com/apache/tika/blob/main/tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-ocr-module/src/test/java/org/apache/tika/parser/ocr/TesseractOCRParserTest.java#L66

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-cad-module/src/main/java/org/apache/tika/parser/dwg/DWGParser.java
```diff
@@ -94,6 +93,14 @@ public Set<MediaType> getSupportedTypes(ParseContext context) {
 
     public void parse(InputStream stream, ContentHandler handler, Metadata metadata,
                       ParseContext context) throws IOException, TikaException, SAXException {
+    	
+    	configure(context);
+    	DWGParserConfig dwgc = context.get(DWGParserConfig.class);
+
+if(!dwgc.getDwgReadExecutable().isEmpty()) {
```
tballison, line 100: Do I remember correctly that the dwg commandline handles a different version of files than our existing parser.  If so, we'll want to branch based on detected mime type, not whether the executable is available.

If this is as intended, please ignore this comment.

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-cad-module/src/main/java/org/apache/tika/parser/dwg/DWGParser.java
```diff
@@ -128,7 +135,7 @@ public void parse(InputStream stream, ContentHandler handler, Metadata metadata,
                 throw new TikaException("Unsupported AutoCAD drawing version: " + version);
         }
 
-        xhtml.endDocument();
+        xhtml.endDocument();}
```
tballison, line 138: formatting.  Did checkstyle let this one through?

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-cad-module/pom.xml
```diff
@@ -36,6 +36,17 @@
       <artifactId>tika-parser-microsoft-module</artifactId>
       <version>${project.version}</version>
     </dependency>
+
+    <dependency>
+    	<groupId>com.fasterxml.jackson.core</groupId>
+    	<artifactId>jackson-core</artifactId>
+	    <version>${jackson.version}</version><!--$NO-MVN-MAN-VER$-->
```
tballison, line 43: Shouldn't need versions for jackson-core or jackson-databind because those are in dependency management in tika-parent.  I'm still in the process of cleaning these up, and there are quite a few cases where I've done exactly this.  This isn't your fault.

tballison (commented):
Overall, this is impressive.  There are a few things that you can reuse from the OCR parser that might make this a bit tighter.

I'm not clear on how this should interact with our existing parser...do they cover the same mimes?  Or do you want to use our current parser only if the dwg executable is not available.

Many, many thanks for this!

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-cad-module/src/main/java/org/apache/tika/parser/dwg/DWGReadParser.java
monkmachine: Added tmpFileOutCleaned to be deleted to make sure it is deleted if we encounter exception

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-cad-module/src/main/java/org/apache/tika/parser/dwg/DWGParser.java
monkmachine: I'm pretty sure DWGRead takes all the versions the current reader does, it only accepts the same mime type as the original DWGParser which is why I've left it here.

tballison: K.  So the logic is if dwgread is available, we'll skip our current code?  Sounds good!

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-cad-module/src/main/java/org/apache/tika/parser/dwg/DWGReadParser.java
monkmachine: @tballison added ProcessUtils.execute instead of using ProcessBuilder manually, also added timeout (default 5mins as I have found some do take some time) to the DWGParserConfig

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-cad-module/src/main/java/org/apache/tika/parser/dwg/DWGParser.java
monkmachine: should now be fixed

nddipiazza:
@tballison @monkmachine 
> Or do you want to use our current parser only if the dwg executable is not available.
I would vote +1 on *use current parser only if the dwg executable is not available.*

nddipiazza:
should we use TestContainers to test this within a docker container to make sure it works?

monkmachine:
> @tballison @monkmachine
> 
> > Or do you want to use our current parser only if the dwg executable is not available.
> 
> I would vote +1 on _use current parser only if the dwg executable is not available._

Yeah that's what I think is the best solution :) Thanks Nicolas for the initial code, hopefully I haven't butchered yours too much. I did manage to find we hold text in "value_text" and "text" fields. MTEXT fields also contain formatting which I will remove (I've done a quick stab at some of the main ones contain within the documents I have but will do them all before I finish the PR)

monkmachine:
> should we use TestContainers to test this within a docker container to make sure it works? or is it sufficient to just run test only if dwgread is installed?

Would this be on the Apache hosted build server? If so would you construct the test as I wouldn't know the path to dwgread?

nddipiazza:
> hopefully I haven't butchered yours too much.

are you kidding me? you're awesome! all looks great to me. 

> Would this be on the Apache hosted build server? 

yes - that's right. 

on the tika pipes module we had simply used the TestContainers java test utility and it is pretty easy. You can see that code for an example

tballison:
>are you kidding me? you're awesome! all looks great to me.
+1 Thank you!

tballison:
> should we use TestContainers to test this within a docker container to make sure it works? or is it sufficient to just run test only if dwgread is installed?

Maybe?  I worry about the build times and network resources to pull the images.  But testing is critical.  Can we may do @Nightly/@Weekly/@ShortlyBeforeTheNextRelease or something?

If we go down this route, we should consider doing the same for tesseract.

monkmachine:
@nddipiazza @tballison 
This looks messy, can you advise a way to clean it up? A better way of doing it? Still think its worth having the comments there?
https://github.com/apache/tika/blob/28ceff420948ede5e18ba10a4f1d6d1751f30b3b/tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-cad-module/src/main/java/org/apache/tika/parser/dwg/DWGReadParser.java#L272-L318

tballison:
> @nddipiazza @tballison This looks messy, can you advise a way to clean it up? A better way of doing it? Still think its worth having the comments there?

OMG, what a mess.  The output, not you.

What I've done before is a regex pattern+matcher that captures the escape sequence first OR then the controls ~/(\\\\)|(\\[A-Z0-9]{1,5})/, capture group(2) (and skip it), append group 1 to tail.

That's a rough answer and probably wrong, but see what you can do.

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-cad-module/src/main/java/org/apache/tika/parser/dwg/AbstractDWGParser.java
tballison: If a user doesn't have dwgread on their system, will this throw a config exception during initialization?

monkmachine: > 

Hi @tballison 
It seems to work ok, I've compiled a version of tika server and if I don't supply the parser config it falls back to the old parser. If the config is incorrect then Tika server fails to start (which is what we want?)

Parser config looks like this:
```
<parser class="org.apache.tika.parser.dwg.DWGReadParser">
	<params>
		<param name="dwgReadExecutable" type="string">G:/libredwg-0.12.5-win64/dwgread.exe</param>
		<param name="cleanDwgReadOutput" type="bool">true</param>
		<param name="cleanDwgReadOutputBatchSize" type="int">10000000</param>
		<param name="dwgReadTimeout" type="long">10</param>
		<param name="cleanDwgReadRegexToReplace" type="string">[^\x20-\x7e]| nan,| nan </param>
		<param name="cleanDwgReadReplaceWith" type="string"/>
	</params>
</parser>

tballison: W00t!

tballison:
> > @nddipiazza @tballison This looks messy, can you advise a way to clean it up? A better way of doing it? Still think its worth having the comments there?
> 
> OMG, what a mess. The output, not you.
> 
> What I've done before is a regex pattern+matcher that captures the escape sequence first OR then the controls ~/(\\)|(\[A-Z0-9]{1,5})/, capture group(2) (and skip it), append group 1 to tail.
> 
> That's a rough answer and probably wrong, but see what you can do.
> 
> The braces...hmmmm... Maybe take a second pass and do the same thing? You can't just add this in the OR ~/{[^}]{0,50}}/ because that'll not correctly process escaped } within the brackets.

I threw together a somewhat working example.  I think there are still some things I'm missing: https://github.com/tballison/tika-addons/blob/main/DWGReadDev/src/test/java/TestRegexCleaners.java

Obv, we'll want to make the patterns static, etc.

monkmachine:
> 



> > > @nddipiazza @tballison This looks messy, can you advise a way to clean it up? A better way of doing it? Still think its worth having the comments there?
> > 
> > 
> > OMG, what a mess. The output, not you.
> > What I've done before is a regex pattern+matcher that captures the escape sequence first OR then the controls ~/(\)|([A-Z0-9]{1,5})/, capture group(2) (and skip it), append group 1 to tail.
> > That's a rough answer and probably wrong, but see what you can do.
> > The braces...hmmmm... Maybe take a second pass and do the same thing? You can't just add this in the OR ~/{[^}]{0,50}}/ because that'll not correctly process escaped } within the brackets.
> 
> I threw together a somewhat working example. I think there are still some things I'm missing: https://github.com/tballison/tika-addons/blob/main/DWGReadDev/src/test/java/TestRegexCleaners.java
> 
> Obv, we'll want to make the patterns static, etc.

Will take a look @tballison , thanks for your help. I've been cleaning up the code to match the checkstyle (which I've only learnt about today) and testing my janky regexes (in the current form) against some documents I have.  Like I said I managed to build Tika Server and check the config was working correctly so been a successful few hours today :) Will take a look at your example tomorrow and hopefully at some point this week find some time to check the stop method on the other pull request. We can then look to create a guide/script on how to install Tika Server as a windows service using Daemon.

monkmachine:
Help! @tballison  @nddipiazza  Any reason why this section would sometimes write extra lines out? On some json files when cleaning up it writes out the file correctly then appends another section which makes the json invalid? I'm getting extra lines in some (very few) files which I can't share unfortunately. I can't for the life of me work out why.... I've even tried removing the replaceAll 

https://github.com/apache/tika/blob/a3e2be9b69de6a0550f1211aa9ff8df96c60f767/tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-cad-module/src/main/java/org/apache/tika/parser/dwg/DWGReadParser.java#L112-L122

monkmachine:
If I read byte by byte (i.e. byte[] bytes = new byte[1];) I get the correct result:
image

If I read with anything other than byte by byte I get added bytes/strings from some other part of the file:
image

It's only doing it on this one json file, every time I can reproduce it every time on this one.


``` @Test
    public void jsonConvert() throws FileNotFoundException, IOException {



    	 try (FileInputStream fis = new FileInputStream("c:\\temp1\\dwgreadout.json");
                 FileOutputStream fos = new FileOutputStream("c:\\temp1\\dwgreadoutClean.json")) {
             byte[] bytes = new byte[1000];
             while (fis.read(bytes) != -1) {
                 byte[] fixedBytes = new String(bytes, StandardCharsets.UTF_8)
                 		
                         //.replaceAll(dwgc.getCleanDwgReadRegexToReplace(), dwgc.getCleanDwgReadReplaceWith())
                         //.replaceAll(" nan ", " 0 ")
                         //.replaceAll(" nan,", " 0,")
                         .getBytes(StandardCharsets.UTF_8);
                 String st = new String(fixedBytes, StandardCharsets.UTF_8);
                 fos.write(fixedBytes, 0, fixedBytes.length);
                 
                 
             }
         } 
    }

monkmachine:
If I use buffer reader I get the correct output but it's slower: 3s vs 10s (it's quite a large file)
```    public void jsonConvert() throws FileNotFoundException, IOException {



    	    //FileInputStream fis = new FileInputStream("c:\\temp1\\dwgreadout.json");
            FileOutputStream fos = new FileOutputStream("c:\\temp1\\dwgreadoutClean.json");
            try (BufferedReader br = new BufferedReader(new FileReader("c:\\temp1\\dwgreadout.json"))) 
            {

                String sCurrentLine;
                while ((sCurrentLine = br.readLine()) != null) 
                {
                	sCurrentLine = sCurrentLine
                	    .replaceAll(" nan ", " 0 ")
                            .replaceAll(" nan,", " 0,") +"\n";
                	fos.write(sCurrentLine.getBytes(), 0, sCurrentLine.getBytes().length);
                }
            
                 //fos.write(fixedBytes, 0, fixedBytes.length);
                 
                 
             }}

tballison:
> If I use buffer reader I get the correct output but it's slower: 3s vs 10s (it's quite a large file)
> 
> ```
> 
> 
> 
>     	    //FileInputStream fis = new FileInputStream("c:\\temp1\\dwgreadout.json");
>             FileOutputStream fos = new FileOutputStream("c:\\temp1\\dwgreadoutClean.json");
>             try (BufferedReader br = new BufferedReader(new FileReader("c:\\temp1\\dwgreadout.json"))) 
>             {
> 
>                 String sCurrentLine;
>                 while ((sCurrentLine = br.readLine()) != null) 
>                 {
>                 	sCurrentLine = sCurrentLine
>                 	    .replaceAll(" nan ", " 0 ")
>                             .replaceAll(" nan,", " 0,") +"\n";
>                 	fos.write(sCurrentLine.getBytes(), 0, sCurrentLine.getBytes().length);
>                 }
>             
>                  //fos.write(fixedBytes, 0, fixedBytes.length);
>                  
>                  
>              }}
> ```

I agree that it's slower, but it is more likely to be correct.  I don't understand the problem as thoroughly as you do, but running regexes against a part of a file at a time is guaranteed to fail depending on where the breaks in the parts are.

tballison:
Can you guarantee that reading per line will be ok on this json-disaster?  If so, that's the way to go.  The other thing is that it may be better to specify the encoding and work on the string instead of bytes.  By better, I only mean less likely to break quietly...there are no perfect solutions for this.

tballison:
Can you tell if they're writing utf8?  Are there any ascii accented data items or non-ascii characters that you can use to figure out what they're default encoding is?

tballison:
If NaN is the only problem, is there any way to tell jackson to be lax?  Maybe something like: https://stackoverflow.com/questions/63516411/convert-nan-value-into-null-when-parsing-json-using-jackson-library

tballison:
No, that probably won't work. Sorry.  If you send me some examples, I can try some things.

monkmachine:
> No, that probably won't work. Sorry. If you send me some examples, I can try some things.

Yeah we'd be ok if Jackson allowed "nan" as well as "NaN" as we could use JsonReadFeature.ALLOW_NON_NUMERIC_NUMBERS.

monkmachine:
> Can you tell if they're writing utf8? Are there any ascii accented data items or non-ascii characters that you can use to figure out what they're default encoding is?

If you can have a scan of the library to work it out?
https://github.com/LibreDWG/libredwg/tree/master/src

monkmachine:
> Can you guarantee that reading per line will be ok on this json-disaster? If so, that's the way to go.
> 
> The other thing is that you'll want to specify the encoding on your reader. It is much better to work with the String than the bytes.

I've got to test it tomorrow with my files but looks to be promising.

monkmachine:
> No, that probably won't work. Sorry. If you send me some examples, I can try some things.
Can't send you examples unfortunately :( 
I did manage to process over 1000's files today with it and compared to a .net libary (CAD .Net) it does look better. I also managed to find a few more formatting's that are output that weren't on the two websites.

PR #605, opened 2022-07-12
TIKA-3815: set GMT timezone for unspecified timezones like drew noakes,

lfcnassif:
Fixes TIKA-3815.

Actually Drew Noakes metadata-extractor library uses **GMT** timezone when timezone is not specified since version 2.8.1, not JVM default:
[fix tests to work on different timezones](https://github.com/drewnoakes/metadata-extractor/commit/1899deac8af55934363183dae605e286bc79afa3)

So I changed the code to reflect that and updated tests. Now `user.timezone` doesn't need to be fixed to UTC anymore to run tests, they should work on any timezone.

I would like some other committer to test this to be sure tests work on a timezone different than mine (GMT-3)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -13,6 +13,9 @@ Release 2.4.2 - ???
    * Fix bug that prevented specification of rendered image type
      via http header in the PDFParser (TIKA-3794).
 
+   * Fix bug causing some Exif dates to be decoded wrongly on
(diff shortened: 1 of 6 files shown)


lfcnassif:
Should this be also applied to 1.x branch or we are just fixing security issues?

lfcnassif:
I did a few more tests and I think this is fine, I'm going to merge.

PR #616, opened 2022-07-18
TIKA-3820 add tika pipes kafka support

nddipiazza:
See https://issues.apache.org/jira/browse/TIKA-3820

Add support for Kafka as a pipe iterator, and Kafka as an emitter. 

(A fetcher would make no sense here, thus no fetcher).


tballison (approved):
This looks great. LGTM!

PR #633, opened 2022-08-01
[TIKA-3825] ForkClient to check for thread interrupted status

TheHound:
ForkClient to check for thread interrupted status when waiting for response. Add test to ForkParserTest to demonstrate issue and fix. Targeting branch_1x


TheHound:
Sorry my dev VM crashed before I could fix this, there is a compilation error in `ForkParserTest` just running another `mvn clean install` to ensure its fixed and I'll push another commit to this branch.

PR #748, opened 2022-10-17
TIKA-3883 Fixes for Parsing DWG files using DWG read

monkmachine:
Fixes for Parsing DWG files using DWG read Parser identified during testing


tballison:
Y, probably good enough. If you find it easy enough to mock up some json and run against that in the future, that'd be useful.  Thank you!

PR #936, opened 2023-02-02
Create dependency reduced pom.xml for tika-parsers-standard-package

gastaldi:
`tika-parsers-standard-package` is a shaded JAR and the pom.xml keeps the transitive dependencies used to generate the shaded JAR

### DIFF
--- a/tika-parsers/tika-parsers-standard/tika-parsers-standard-package/pom.xml
+++ b/tika-parsers/tika-parsers-standard/tika-parsers-standard-package/pom.xml
@@ -382,7 +382,7 @@
             </goals>
             <configuration>
               <createDependencyReducedPom>
-                false
+                true
               </createDependencyReducedPom>
               <artifactSet>
                 <includes>


sberyozkin:
Hi Tim, please review this PR from my colleague George

tballison:
There are likely several other places we should do this.  I've opened https://issues.apache.org/jira/browse/TIKA-3966 to track this work.  Thank you @gastaldi!

tballison:
@gastaldi and @sberyozkin  I should have foreseen the problem.

This is causing problems for users in our 2.8.0-rc1 when they try to import tika-parsers-standard-package, the transitive dependencies are not getting pulled in.  Any recommendations?

cc @jnioche

gastaldi:
Aren't the binaries from the transitive dependencies are already included in this JAR?

tballison:
They are not.

tballison:
Y, that works.  Let me see if there are any objections on https://issues.apache.org/jira/browse/TIKA-4038

gastaldi:
@tballison I created https://github.com/apache/tika/pull/1130 that does that. Build seems okay

PR #975, opened 2023-02-22
fix TIKA-3970

nddipiazza:
- start correctly handling the onlyLatestRevision option by detecting parent property id of type ElementChildNodesOfVersionHistory and do not print those if that flag is set.
- need to prevent File Node references to the same text from being printed out in the xml stream multiple times.
- improve PropertyValue to have a toString so that debugging in IDE is easier.


nddipiazza:
@tballison please review for fix of https://issues.apache.org/jira/browse/TIKA-3970

tballison:
@nddipiazza Many thanks!

PR #985, opened 2023-02-24
[TIKA-3979] OneNoteParser - Improve performance for deserialization

davidxie-glean:
Context
---
CPU profiling of OneNoteParser shows that a lot of cpu time is spent creating the BitSet in `BitReader`. Reusing the byte array reference will prevent the need to make a copy of the byte array.

Template
---

* the issue ID (`TIKA-XXXX`)

### DIFF
--- a/tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-microsoft-module/src/main/java/org/apache/tika/parser/microsoft/onenote/fsshttpb/util/BitReader.java
+++ b/tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-microsoft-module/src/main/java/org/apache/tika/parser/microsoft/onenote/fsshttpb/util/BitReader.java
@@ -38,7 +38,7 @@ public class BitReader {
     /**
      * A byte array which contains the bytes need to be read.
      */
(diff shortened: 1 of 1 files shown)


tballison:
Do you have any performance comparisons before/after?  Thank you for this PR!

nddipiazza:
i ran a few hundred results and i see literally no difference in performance.

nddipiazza:
Never mind. i was testing wrong format. we are not concerned only with the alternative packaging format, and my previous test didn't.

100 documents parsed previously in 10252 ms
now 100 documents parsed previously in 1062 ms

so yeah. big win.

apismensky:
I was going to submit this issue last week. 
My observation was similar - lots of overhead around BitSet - mem allocations / cpu. 
We switched from tika 1.27 to 2.7.0 
For one of the files we saw the difference: 
Extraction took: 2199 ( tika 1.27) vs
Extraction took: 27010 ( tika 2.7.0) 

Both in ms, so it is more than 10 times slower.
Original file size is 50.5 Mb

apismensky:
Confirming with my file: 
Before fix: 26844 ms
After fix: 692 ms
Yay yay! 
@nddipiazza thanks for fixing!

PR #1522, opened 2023-12-26
Fix html parsing logic when body contains script tag

yakivy:
Hey there. I spotted that public signup to Tika Jira is disabled, so I'm not sure what is the correct flow for issues without a ticket, putting all related info here...

### Issue description
`org.apache.tika.parser.html.JSoupParser` parser ignores `body` content after the `script` tag. Check `org.apache.tika.parser.html.HtmlParserTest#testScriptInBody` test for details.

### Fix description
The main reason of such behaviour is that `#data` tag tail decrements `org.apache.tika.parser.html.HtmlHandler.bodyLevel`, when `#data` tag head is considered a content node and doesn't invoke `startElement` logic at all. As a fix I mirrored head logic for the data node to the tail and returned `FilterResult.CONTINUE` early.


tballison:
Thank you @yakivy!

PR #1589, opened 2024-02-12
[TIKA-4194] Fix for unrecognized pkcs12 keystores

Lonzak:
Fixes the issue that some pkcs12 keystores are not correctly detected. Tested with 157x p12 keystores (kindly provided by [redhat](https://github.com/redhat-qe-security/keyfile-corpus/tree/master).)


Lonzak:
It would appreciated if the change could go into 2.9.X ([branch_2x](https://github.com/apache/tika/tree/branch_2x))

PR #1698, opened 2024-03-28
TIKA-4229 - Add a Microsoft Graph Fetcher

nddipiazza:
initial attempt to add microsoft graph fetcher


tika-pipes/tika-fetchers/tika-fetcher-microsoft-graph/src/main/java/org/apache/tika/pipes/fetchers/microsoftgraph/MicrosoftGraphFetcher.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    @Override
+    public InputStream fetch(String fetchKey, Metadata metadata) throws TikaException, IOException {
+        int tries = 0;
+        Exception ex;
+        do {
+            try {
+                long start = System.currentTimeMillis();
+                String[] fetchKeySplit = fetchKey.split(",");
```
bartek, line 118: Download URLs provided for each drive item seem to be unique (we can confirm this) and would prevent this specific nature of a fetch key being provided. This would simplify the call to the fetcher as well.

Thoughts?

bartek: I'm stepping back on this comment. After reviewing the responses from the MS Graph API, there is no simple (without creating additional resources in the API) to get a long-living download URL. The URLs it provides are short lived (typically one hour) with a temp auth token. This is not ideal for Tika

The drive item ID approach here I believe is the right one. I'll be testing it soon.

PR #1712, opened 2024-04-05
Tika 4237 add jwt authentication ability to the http fetcher

nddipiazza:
Add new parameters in the HttpFetcher so that it can properly handle JWT bearer token generation.

JWT generation via secret key and private key.

- String jwtIssuer
- String jwtSubject
- int jwtExpiresInSeconds

and then you can specify:

- String jwtSecret
or
- String jwtPrivateKeyBase64

When specified, the authorization header will be added to all requests.


tika-pipes/tika-fetchers/tika-fetcher-http/src/main/java/org/apache/tika/pipes/fetcher/http/jwt/JwtGenerator.java
```diff
@@ -0,0 +1,64 @@
...
+import com.nimbusds.jose.JOSEException;
+import com.nimbusds.jose.JWSAlgorithm;
+import com.nimbusds.jose.JWSHeader;
+import com.nimbusds.jose.JWSSigner;
+import com.nimbusds.jose.crypto.MACSigner;
+import com.nimbusds.jose.crypto.RSASSASigner;
+import com.nimbusds.jwt.JWTClaimsSet;
+import com.nimbusds.jwt.SignedJWT;
+
+public class JwtGenerator {
```
bartek, line 16: One thing we need to support is the `qsh` claim addition for Atlassian JWTs, see: https://developer.atlassian.com/cloud/jira/platform/understanding-jwt-for-connect-apps/

This `qsh` claim takes the incoming request URL (ie download URL) and computes a hash from it. My thinking is we could send the calculated value of the `qsh` claim as part of FetchAndParse, and it would be amended to the claims during token generation. Every request would of course create a new JWT as this `qsh` claim value would be different.

Is that possible in this workflow?

nddipiazza: hi we talked about this offline. just posting here. this was very hairy and involved a lot of code to be copied from the atlassian library for this to have worked. therefore, anyone who needs this just needs to create a simple Fetcher, build and add to the classpath of tika-grpc, then they can use this feature. but it's custom dev thus not doing it here.

PR #1753, opened 2024-05-10
TIKA-4252: add request metadata

nddipiazza:
add request metadata

### DIFF
--- a/tika-core/src/main/java/org/apache/tika/pipes/PipesServer.java
+++ b/tika-core/src/main/java/org/apache/tika/pipes/PipesServer.java
@@ -455,33 +455,33 @@ public class PipesServer implements Runnable {
         }
     }
 
-    protected MetadataListAndEmbeddedBytes parseFromTuple(FetchEmitTuple t, Fetcher fetcher) {
-        FetchKey fetchKey = t.getFetchKey();
+    protected MetadataListAndEmbeddedBytes parseFromTuple(FetchEmitTuple fetchEmitTuple, Fetcher fetcher) {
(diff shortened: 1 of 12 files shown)


tika-core/src/main/java/org/apache/tika/pipes/PipesServer.java
```diff
@@ -455,33 +455,33 @@ private Fetcher getFetcher(FetchEmitTuple t) {
         }
     }
 
-    protected MetadataListAndEmbeddedBytes parseFromTuple(FetchEmitTuple t, Fetcher fetcher) {
-        FetchKey fetchKey = t.getFetchKey();
+    protected MetadataListAndEmbeddedBytes parseFromTuple(FetchEmitTuple fetchEmitTuple, Fetcher fetcher) {
+        FetchKey fetchKey = fetchEmitTuple.getFetchKey();
+        Metadata fetchResponseMetadata = new Metadata();
```
tballison, line 460: The metadata that goes in the fetchemittuple was envisioned to be user-injected metadata that passed through the parse process and was emitted (provenance metadata).

I think we need to put both metadatas on the fetchemittuple.

tballison: @nddipiazza any chance you can revert this in main so that we have a working build? Thank you!

nddipiazza: shoot i didn't realize i was deplying broken builds! reverted. i'll make this change and make a new pr

PR #1774, opened 2024-05-23
TIKA-4252 fetch tuple metadata

nddipiazza:
Add ability to add Tika Fetch Metadata


nddipiazza:
oops not quite right - need to sync up with @tballison to make sure i'm covering his needs and not just my own

PR #1832, opened 2024-06-26
TIKA-4181 move grpc into tika-server and use shade plugin to bundle it

nddipiazza:


tballison:
Is there any chance you'd consider using the maven dependencies plugin to create a lib directory with all dependencies so that we don't have to use the shade plugin?

We use shading quite a bit in Tika. I've seen other projects go the lib/* route (nutch, solr, etc), and it seems a bunch cleaner and less errorprone than shading.

I was contemplating moving tika-server to this model for 3.x.

Not sure what fellow devs think?

nddipiazza:
we closed this for 1853

PR #1872, opened 2024-08-03
TIKA-4290 Attempt to fix error "instance of Class compared with String"

dk2k:
TIKA-4290 Attempt to fix error "instance of Class compared with String"

### DIFF
--- a/tika-serialization/src/main/java/org/apache/tika/serialization/pipes/JsonFetchEmitTuple.java
+++ b/tika-serialization/src/main/java/org/apache/tika/serialization/pipes/JsonFetchEmitTuple.java
@@ -109,7 +109,7 @@ public class JsonFetchEmitTuple {
             String superClassName = readVal(TikaJsonSerializer.SUPER_CLASS_KEY, obj, className, false);
             try {
                 Class clazz = Class.forName(className);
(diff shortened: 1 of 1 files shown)


THausherr:
I made a comment at the bottom of TIKA-4252 because even with the correction, that line of code doesn't make sense to me, the alternative isn't needed, the result is always the same.

tballison:
There's redundancy in the serialization/deserialization of ParseContext, where the key name is the super class. I wanted to reuse TikaJsonSerializer#serializeObject which needs to encode the instantiation class and the superclass because normally the superclass is not defined. Let me try to clean this up.

This is the redundancy that is part of ParseContext:

```
"parseContext": {
    "org.apache.tika.metadata.filter.MetadataFilter": {
      "_class": "org.apache.tika.metadata.filter.CompositeMetadataFilter",
      "_super_class": "org.apache.tika.metadata.filter.MetadataFilter",
      "filters": [
        {
          "_class": "org.apache.tika.metadatafilters.SomeFilter",
```

THausherr:
Tim will fix it in https://issues.apache.org/jira/browse/TIKA-4294 , so I guess I should close this PR. Thanks for finding this!

PR #1873, opened 2024-08-03
TIKA-4290 Attempt to fix error "String.format() was called with 1 arg…

dk2k:
…, but expects none". Not sure if we want output int type as %d somewhere in the exception's message


THausherr:
Yes please do, replace RuntimeException with %d and if you can, please squash the commits.

dk2k:
@THausherr  Squash is an alternative to merge, but both are blocked now - the check wasn't executed. Furthermore, I doubt that I have permission to squash or merge

THausherr:
Ignore my comment, I just saw that I can do it when merging. I'll do it as soon as the check is done.

PR #1875, opened 2024-08-04
TIKA-4290 Removed unused collections and a method call

dk2k:
TIKA-4290 Removed unused collections and a method call


THausherr:
I suspect this is unfinished code, except maybe the 3rd file. ping @tballison  The first file has "TIKA-3547 -- first draft of a new ExternalParser" in its history.

tballison:
Looks like checkstyle issues in tika-core.

I concur on the second two files. The first one, Param, that's probably a vestigial bit of code for handling maps that was based on `valueStrings` for lists. I agree that should be removed.

THausherr:
re checkstyle - @dk2k could you check if the import of LinkedHashMap can be removed?

PR #1888, opened 2024-08-05
TIKA-4290 Used try-with-resources for some Closeables. 

dk2k:
TIKA-4290 Prevent resource leaks


THausherr:
A space is missing before "{" in GeographicInformationParser.java. It also complains that "CRLF line endings are prohibited" but this might be just my problem.

PR #1960, opened 2024-09-18
[TIKA-4312] Make audio/mp4 a subclass of video/mp4

c960657:
https://issues.apache.org/jira/browse/TIKA-4312

tika-mimetypes.xml currently defines these two <sub-class-of> ancestor chains:

- audio/mp4 → application/quicktime.
 - video/mp4 → video/quicktime → application/quicktime.

Based on [RFC 4337, section 2](https://datatracker.ietf.org/doc/html/rfc4337#section-2), it is my understanding that audio/mp4 is a special case of video/mp4. I thus suggest changing the ancestor chain of audio/mp4 to 

- audio/mp4 → video/mp4 → video/quicktime → application/quicktime


tballison:
Thank you!

PR #1961, opened 2024-09-20
[TIKA-4309] detailed Mach-O in tika-mimetypes.xml

alexey-pelykh:
A follow-up to #1947: Support 32-bit and 64-bit Mach-O files and detect their type

cc @tballison

PR #1963, opened 2024-09-21
[TIKA-4309] Mach-O Universal in tika-mimetypes.xml

alexey-pelykh:
A follow-up to #1947 & #1961 & #1962: Mach-O universal + optimized Mach-O

cc @tballison


alexey-pelykh:
Probably I'm testing in a wrong way locally as I didn't see a failing test on my side

alexey-pelykh:
Updated to make it a non-breaking change

tballison:
Aren't the magic bits at 4096 important for jnilib? Or what is the difference between what we were detecting as `x-java-jnilib` and mach-O and its variants?

```
  <mime-type type="application/x-java-jnilib">
    <_comment>Java Native Library for OSX</_comment>
    <magic priority="50">
      <match value="0xcafebabe" type="string" offset="0">
        <match value="0xfeedface" type="string" offset="4096"/>
        <match value="0xfeedfacf" type="string" offset="4096"/>
        <match value="0xcefaedfe" type="string" offset="4096"/>
        <match value="0xcffaedfe" type="string" offset="4096"/>
      </match>
    </magic>
    <glob pattern="*.jnilib"/>
  </mime-type>
```

alexey-pelykh:
> Aren't the magic bits at 4096 important for jnilib? Or what is the difference between what we were detecting as x-java-jnilib and mach-O and its variants?

Yes and no. There's no guarantee they are at 4096 offset, that's empiric - at least based on what I know as of now. The offset should be read from the file itself - that's why I've asked about dynamic offset. An top of that, we should check that it's a library, not that it's a Mach-O, so technically those checks aren't ideal either.

So far we're mimicking the `file` behavior that was also used by this:

> This might be useful: https://github.com/gabriel-vasile/mimetype/pull/53/files#diff-db84ad344d2bf91ea9533dbaa5974e81940b053644665471594fb2eedd34e7daR19 ?

tballison:
K, that's what I was trying to figure out -- the diff between what we were calling a jnilib (https://issues.apache.org/jira/browse/TIKA-1169 for the original work on `jnilib`) and mach-o

PR #1993, opened 2024-10-16
TIKA-4309 -- some mods

tballison:
@alexey-pelykh let me know what you think. I probably should have recommended these changes before merging into main... oh well.

I moved the parser into the code module. 
I got rid of some of the `var`s so that we had explicit int vs long. 
I used IOUtils.skipFully().

If you can recommend an upper bounds check in the TODO below, that'd be great.

Thank you, again!


tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-code-module/src/main/java/org/apache/tika/parser/executable/UniversalExecutableParser.java
```diff
@@ -119,46 +118,62 @@ public void parseMachO(XHTMLContentHandler xhtml, EmbeddedDocumentExtractor extr
...
             var perArchMetadata = new Metadata();
             var tikaInputStream = TikaInputStream.get(perArchMachO, perArchMetadata);
             if (extractor.shouldParseEmbedded(perArchMetadata)) {
-                extractor.parseEmbedded(tikaInputStream, handler, perArchMetadata, true);
+                extractor.parseEmbedded(tikaInputStream, xhtml, perArchMetadata, true);
             }
         }
     }
+
+    private static class Pair {
```
alexey-pelykh, line 157: Why not to use `org.apache.commons.lang3.tuple.Pair` since we're depending on that package anyhow?

tballison: Y, probably should... I was thinking about that. Strictly, the package parser module depends on it. The code parser module hadn't.

alexey-pelykh: Other that this, looks reasonable!

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-code-module/src/main/java/org/apache/tika/parser/executable/UniversalExecutableParser.java
```diff
@@ -48,6 +49,9 @@ public class UniversalExecutableParser implements Parser {
     private static final Set<MediaType> SUPPORTED_TYPES =
             Collections.singleton(MediaType.application("x-mach-o-universal"));
 
+    private static final int MAX_ARCHS_COUNT = 1000;
```
tballison, line 52: @alexey-pelykh are these reasonable limits?

alexey-pelykh:
I think they are more than generous for real-life usage.

PR #2028, opened 2024-10-28
[TIKA-4337] Fix XPS regressions

ruwi-next:
Fixes exceptions raised and poor content parsing for XPS files.

These fixes in the first commit are for the exceptions:
- NPE when indices attribute is not present
- Parse errors when parsing indices string, we only need the advance so I've simplified this parsing to just grab that float which is in the second comma separated value

In the second commit these are improvements to text extractions:
- Don't assume glyph runs appear in vertical order
- Check for VisualBrush transform which moves child canvases
- Use the entire canvas stack for keying the canvas map instead of just the last one
- Split GlyphRuns with large gaps into two glyph runs because in some cases another glyph run will be positioned in that gap meaning the extracted text order needs to be different.
- Fix text direction assumed to be RTL when not present


tballison:
Thank you @ruwi-next!

PR #2085, opened 2025-01-03
TIKA-4365 : feat : Support Detection for AAB Bundle file

subbudvk:
TIKA-4365 : Support added for Android AAB bundle file. 

Detection is through AndroidManifest.xml similar to APK files in JarDetector. As per Android Developer Documentation this file is seen in manifest directory. 

https://developer.android.com/guide/app-bundle/app-bundle-format


subbudvk:
@tballison  you are right. The Jar detector decides content based on the presence of MANIFEST.MF. The android package bundle will also have MANIFEST.MF, so to guarantee the order and JarDetector doesn't return jar based on MANIFEST.MF  and since other android types are already handled in JarDetector, I changed here. 

Let me know if you have comments.

tballison:
I'll merge this as soon as the checks pass. On a follow-on PR, if you have a chance, would you be able to create/find an ASL 2.0-compatible test file for a unit test?

subbudvk:
I will add tests for this @tballison. Thanks for the help!

tballison:
I just merged into main. I'll fix checkstyle now.

PR #2098, opened 2025-01-13
[TIKA-4303] Handle OneNotePropertyEnum.CachedTitleString as RichEditTextUnicode

sunluman:
Fixes https://issues.apache.org/jira/browse/TIKA-4303

The issue of garbled text is caused by `OneNotePropertyEnum.CachedTitleString` not being correctly parsed. It should be parsed using `handleRichEditTextUnicode`.

As for why versions `2.7.0` and earlier did not encounter garbled text, I believe it was due to a previously erroneous line of code:
```java
if (options.getUtf16PropertiesToPrint().contains(propertyValue.propertyId))
```

This line caused the parsing of OneNote files to never append the parsed content of `OneNotePropertyEnum.ImageFilename, OneNotePropertyEnum.Author, and OneNotePropertyEnum.CachedTitleString` to the xhtml.

However, when parsing `OneNotePropertyEnum.RichEditTextUnicode`, the logic for only parsing the latest version’s content was not added. As a result, the files appeared to be successfully parsed and without garbled text, but in reality, CachedTitleString was never parsed.

I only fixed the bug in the issue where the title in the uploaded file was not parsed. During the testing process, I also discovered the following issues:
- Non-rich text content is not checked for the latest version, so when the content is TextExtendedAscii, it is still parsed repeatedly.
- Dates are not parsed.
- Chinese (or other non-Ascii characters? i'm not sure) characters in the content are not parsed.

I am not sure whether to create a new issue before proceeding with these fixes, so these issues have not been addressed in this PR.

### DIFF
--- a/tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-microsoft-module/src/main/java/org/apache/tika/parser/microsoft/onenote/OneNoteTreeWalker.java
+++ b/tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-microsoft-module/src/main/java/org/apache/tika/parser/microsoft/onenote/OneNoteTreeWalker.java
@@ -393,7 +393,8 @@ class OneNoteTreeWalker {
     private boolean propertyIsBinary(OneNotePropertyEnum property) {
         return property == OneNotePropertyEnum.RgOutlineIndentDistance ||
                 property == OneNotePropertyEnum.NotebookManagementEntityGuid ||
-                property == OneNotePropertyEnum.RichEditTextUnicode;
+                property == OneNotePropertyEnum.RichEditTextUnicode ||
+                property == OneNotePropertyEnum.CachedTitleString;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tballison:
Thank you for opening this and looking deeply into the code. Is there any way to create a unit test for this issue? Do any of our current unit test files show what this fixes? This is not required...we do what we can.

Thank you, again, and y, fixes to the other issues would be great.

tballison:
Thank you! I'd prefer separate issues, but whatever it takes. :rofl:

sunluman:
Sure, if you create a new issue in JIRA (I don’t have an account, so I can’t create one🤣), I’d be happy to work on resolving the issue with reading OneNote content.

THausherr:
You can apply for an account. The secret of getting an account is to enter a small meaningful and specific text, e.g. mention this PR. https://selfserve.apache.org/jira-account.html

sunluman:
Thank you for the reminder, I have submitted an account apply.

tballison:
@sunluman is this ready to go?

nddipiazza:
@sunluman can you produce an example or unit test?

nddipiazza (changes requested):
need a unit test to prove it works.

sunluman:
Sorry, I overlooked this unit test. I will add it right away.

sunluman:
@tballison  @nddipiazza The unit test is ready. This test only performs a simple check for garbled characters in the title.

tballison (approved):

tballison:
@nddipiazza wdyt?

nddipiazza:
@tballison looks great! sorry for the delay sir

PR #2125, opened 2025-02-12
TIKA-4384 : Add ZipX in DefaultMimeTypes

subbudvk:
ZipX is a new Zip extension, the magic works, but name based detection returns application/octet-stream because it is not known to tika-mimetypes.xml


tballison:
Do you happen to know if commons-compress is able to handle zipx files? If not, we should consider a different mime type maybe so that those files don't get routed to commons-compress and then error out.

Any chance you can share a small example file?

Thank you for this!

PR #2126, opened 2025-02-13
TIKA-4385 : fix : GDAL process output read in another thread

lsliwko:
A fix for GDALParser deadlock while reading gdalinfo output.


Illidian4368 (commented):
1

THausherr:
Why is `startFor()` returning an array but only the first element is used?

THausherr:
There's a checkstyle error. Run it locally and look into the file `checkstyle-result.xml` in the target directory of the failing subproject.

THausherr:
Now we have a new problem, but it's not related to your code.
https://github.com/netplex/json-smart-v2/issues/232

THausherr:
Please take the latest commit I did in the trunk

tballison:
I'm sorry for showing up late to this PR. Any chance this could be of use: https://github.com/apache/tika/blob/main/tika-core/src/main/java/org/apache/tika/utils/ProcessUtils.java ?

THausherr:
Yes that would be better, to avoid double implementations.

THausherr:
LGTM

tballison:
Thank you @lsliwko and @THausherr . I'll cherry-pick into 3.x in the next day or two.

PR #2186, opened 2025-04-10
[MINOR] mark some fields as final

pjfanning:
I was looking at some Tika code with the vague idea of copying some of the docx extractor code to Apache POI.
My IDE highlighted that these fields could be final.


tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-microsoft-module/src/main/java/org/apache/tika/parser/microsoft/ooxml/xwpf/XWPFStylesShim.java
```diff
@@ -43,9 +43,9 @@ public class XWPFStylesShim {
     /**
      * Empty singleton to be used when there is no style info
      */
-    public static XWPFStylesShim EMPTY_STYLES = new EmptyXWPFStyles();
+    public static final XWPFStylesShim EMPTY_STYLES = new EmptyXWPFStyles();
```
pjfanning, line 46: this might be dangerous - as the field is public - someone could set the value to a non empty set of styles

PR #2280, opened 2025-07-14
[TIKA-4457] Updating Automatic-Module tag for cad module.

jdeolive:
The original value appears to have been copied from the "code" module. Compiling an upstream project complains of a conflict with the code module.


tballison:
Thank you!

PR #2553, opened 2026-01-27
TIKA-4630 -- Use name from gzip metadata if available, use name as internal path

iachimoe:
A previous change related to the same JIRA means that INTERNAL_PATH is set using the metadata name from a gzip file. However, many gzips don't have this data. Also other archives like bz2 won't have the data. This PR does two things (1) gets the RESOURCE_NAME from the gzip metadata if possible (a change from existing behaviour) and (2), in the absence of a name in the gzip metadata (due to it not being there in a gzip, or another format such as bzip being used), sets the INTERNAL_PATH to be the same as RESOURCE_NAME


tika-parsers/tika-parsers-standard/tika-parsers-standard-package/src/test/java/org/apache/tika/parser/RecursiveParserWrapperTest.java
```diff
@@ -174,6 +174,16 @@ public void testTarball() throws Exception {
                 "/test-documents.tar"), actualEmbeddedPaths);
     }
 
+    @Test
+    public void testTarballWithoutGzipNameMetadata() throws Exception {
+        List<Metadata> list = getRecursiveMetadata("test-documents-no-name-metadata.tgz");
+        Metadata last = list.get(list.size() - 1);
+        String internalPath = last.get(TikaCoreProperties.INTERNAL_PATH);
```
tballison, line 181: The point of internal_path is to store what the file contained about the internal path of a resource. This metadata field should tell the user "this was the path that was literally stored in the container file. Tika did no guesswork here".

PR #2773, opened 2026-04-17
4x-reg-sax-fixes

tballison:

* the issue ID (`TIKA-XXXX`)


utafrali (commented):
The overall direction is correct: these changes make Tika more resilient against truncated and malformed OOXML/ZIP files by converting hard failures into metadata warnings. The main blocker is a real NPE bug in the OPCPackageDetector fallback path where a missing `[Content_Types].xml` entry would silently crash rather than be caught by the `IOException` handler. The other issues (FQCN style, `continue` after iterator failure, no tests) are smaller but worth addressing before merge.

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-microsoft-module/src/main/java/org/apache/tika/detect/microsoft/ooxml/OPCPackageDetector.java
```diff
@@ -281,7 +281,23 @@ public MediaType detect(ZipFile zipFile, TikaInputStream stream) throws IOExcept
...
+            //fall through to [Content_Types].xml fallback below
+        }
+        // POI may have failed (caught above) OR returned null because the
+        // rels were malformed and POI silently produced an empty relationship
+        // collection. Either way, fall back to parsing [Content_Types].xml
+        // directly — same approach as the streaming detector.
+        if (type == null) {
+            try (InputStream contentTypesStream =
+                         zipEntrySource.getInputStream(
+                                 zipEntrySource.getEntry("[Content_Types].xml"))) {
```
utafrali: **NPE risk:** `zipEntrySource.getEntry("[Content_Types].xml")` returns `null` when the entry is absent. Passing that `null` to `getInputStream()` will throw a `NullPointerException`, which the `catch (IOException ignore)` block below won't catch, so the exception propagates up and defeats the entire fallback.

Fix: assign the entry first, check it before opening the stream:
```java
ZipArchiveEntry ctEntry = zipEntrySource.getEntry("[Content_Types].xml");
if (ctEntry != null) {
    try (InputStream contentTypesStream = zipEntrySource.getInputStream(ctEntry)) {
        type = parseOOXMLContentTypes(contentTypesStream);
    } catch (IOException ignore) {
        //swallow
    }
}
```

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-microsoft-module/src/main/java/org/apache/tika/parser/microsoft/ooxml/XSSFExcelExtractorDecorator.java
```diff
@@ -142,24 +142,55 @@ protected void buildXHTML(XHTMLContentHandler xhtml)
...
+        } catch (OpenXML4JException | RuntimeException e) {
+            throw new IOException(e);
+        }
+        // Styles and shared strings are optional — if either part is missing or
+        // unreadable, log to metadata and continue with degraded extraction.
+        try {
+            stylesShim = new XSSFStylesShim(xssfReader.getStylesData(), parseContext);
+        } catch (Exception e) {
+            metadata.add(TikaCoreProperties.TIKA_META_EXCEPTION_WARNING,
+                    org.apache.tika.utils.ExceptionUtils.getStackTrace(e));
```
utafrali: The new catch blocks use fully-qualified names (`org.apache.tika.utils.ExceptionUtils`, `org.apache.tika.exception.WriteLimitReachedException`) while the rest of the file uses short names via imports. This pattern appears three times across the two new catch blocks. Add the missing imports at the top of the file and use the short names, which matches the existing code style.

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-microsoft-module/src/main/java/org/apache/tika/parser/microsoft/ooxml/XSSFExcelExtractorDecorator.java
```diff
@@ -142,24 +142,55 @@ protected void buildXHTML(XHTMLContentHandler xhtml)
...
-            try (InputStream stream = iter.next()) {
+            InputStream nextStream;
+            try {
+                nextStream = iter.next();
+            } catch (RuntimeException e) {
+                // POI can throw POIXMLException for missing sheet parts (e.g.,
+                // truncated workbook references a sheet that isn't in the zip).
+                // Skip the sheet but keep what we've already extracted.
+                metadata.add(TikaCoreProperties.TIKA_META_EXCEPTION_WARNING,
+                        org.apache.tika.utils.ExceptionUtils.getStackTrace(e));
```
utafrali: When `iter.next()` throws a `RuntimeException`, the `continue` returns to the top of the loop where `iter.hasNext()` is called again. If POI's iterator leaves `hasNext()` returning `true` even after the failed `next()` call (e.g., the internal index didn't advance), this becomes an infinite loop, spinning through the same broken sheet repeatedly.

Using `break` instead of `continue` is the safer choice here. If `next()` failed, the iterator state is undefined and abandoning iteration is preferable to risking a hang. Any sheets already processed are preserved in the output.

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-pkg-module/src/main/java/org/apache/tika/parser/pkg/ZipParser.java
```diff
@@ -374,12 +375,22 @@ private void parseStreamEntries(ZipArchiveInputStream zis, Metadata metadata,
             throws TikaException, IOException, SAXException {
 
         try {
-            ArchiveEntry entry = zis.getNextEntry();
-            while (entry != null) {
+            ArchiveEntry entry;
+            while (true) {
+                try {
+                    entry = zis.getNextEntry();
+                } catch (java.util.zip.ZipException ze) {
```
utafrali, line 382: No test resource exercises the `ZipException` path added here. The same applies to the Excel and Word robustness fixes in this PR. Consider adding a few malformed test files (a truncated ZIP, a ZIP-based XLSX with a missing sheet part, a DOCX with corrupt body XML) under `src/test/resources` and corresponding test cases. Without them, future regressions in this error-handling logic won't be caught automatically.

tballison: Please open a pr with asl 2.0 friendly examples. I'm ok skipping new test docs for this one catch.

PR #2791, opened 2026-04-27
TIKA-4721: Fix TOCTOU race in SharedServerManager port assignment

nddipiazza:
## Summary

Fixes the intermittent `testGracefulShutdown` failure in the **main jdk17 windows build (multi-locale)** CI job (tr_TR.UTF-8 / de_DE.UTF-8).

The failure was **not** locale-related despite appearing only in that job. The root cause was a TOCTOU (time-of-check-time-of-use) race in `SharedServerManager.startServer()`:

1. A `ServerSocket` was opened to find a free port → then **closed**
2. The child process was started and told to bind to that same port
3. Between step 1 and step 2, another process (or the OS in TIME_WAIT state, which is far more common on slow Windows runners) could grab the port

When this happened, the child process failed to bind, never printed `READY:{port}`, and the parse on the next iteration returned a non-success result.

## Changes

**`SharedServerManager`**
- Remove the TOCTOU port probe (open/close ServerSocket just to learn a port)
- Pass `TIKA_PIPES_PORT=0` to the child process
- `waitForServerReady()` now returns the actual port read from the `READY:{port}` signal and sets `serverPort` accordingly

**`PipesServer`**
- In `runSharedMode()`, print `READY:{serverSocket.getLocalPort()}` instead of `READY:{port}` so the actual ephemeral port assigned by the OS is reported

## Review Focus Areas

- `SharedServerManager.startServer()` — TOCTOU removal
- `SharedServerManager.waitForServerReady()` — now returns int (actual port)
- `PipesServer.runSharedMode()` — `getLocalPort()` instead of `port`

## Critical Files

- `tika-pipes/tika-pipes-core/src/main/java/org/apache/tika/pipes/core/SharedServerManager.java`
- `tika-pipes/tika-pipes-core/src/main/java/org/apache/tika/pipes/core/server/PipesServer.java`

## Testing Instructions

```bash
cd tika-pipes/tika-pipes-integration-tests
mvn test -Dtest=SharedServerModeTest#testGracefulShutdown
```

The existing CI job **main jdk17 windows build (multi-locale)** should stop failing intermittently once this merges.

## Review Checklist

- [x] No functional behavior change — shared server still starts on a free port, clients connect to the same port
- [x] Removes a latent race condition that was causing intermittent CI failures
- [x] No new dependencies or test infrastructure changes needed

## Potential Concerns

The `READY:{port}` stdout protocol is internal — only `SharedServerManager` reads it. No external consumers are affected by this change.

Closes TIKA-4721

### DIFF
--- a/tika-pipes/tika-pipes-core/src/main/java/org/apache/tika/pipes/core/SharedServerManager.java
+++ b/tika-pipes/tika-pipes-core/src/main/java/org/apache/tika/pipes/core/SharedServerManager.java
@@ -334,7 +334,11 @@ public class SharedServerManager implements ServerManager {
                     String line = reader.readLine();
                     if (line != null && line.startsWith("READY:")) {
                         String portStr = line.substring("READY:".length()).trim();
-                        return Integer.parseInt(portStr);
+                        int port = Integer.parseInt(portStr);
+                        if (port <= 0 || port > 65535) {
+                            throw new IOException("Server reported invalid port: " + port);
+                        }
+                        return port;
                     }
                 } else {
                     // No data available, sleep briefly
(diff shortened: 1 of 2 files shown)


tballison:
LGTM. May want to add a reasonability check on the port > 0 && < 65535? Otherwise, great catch!

PR #2797, opened 2026-05-01
TIKA-4722: Add parse_context_json to FetchAndParseRequest for per-request ParseContext configuration

nddipiazza:
## Summary

Adds a generic `parse_context_json` field (string, field 5) to `FetchAndParseRequest` that allows callers to configure any registered ParseContext component on a per-request basis.

This is a more flexible replacement for a `handler_type`-only approach: any component in the parse-context registry can be overridden at request time, not just the content handler type.

Resolves: [TIKA-4722](https://issues.apache.org/jira/browse/TIKA-4722)

## Changes

- **`tika.proto`**: Add `string parse_context_json = 5;` to `FetchAndParseRequest`
- **`TikaGrpcServerImpl.java`**: Iterate the JSON fields and call `parseContext.setJsonConfig(componentName, valueJson)` for each entry when the field is non-empty
- **`HandlerTypeTest.java`** (e2e): New test verifying HTML vs TEXT output via `parse_context_json`

## Usage Example

```json
{
  "fetch_key": "my-doc.html",
  "fetcher_id": "myFetcher",
  "parse_context_json": "{\"basic-content-handler-factory\": {\"type\": \"HTML\"}}"
}
```

Available `basic-content-handler-factory` types: `TEXT`, `HTML`, `XML`, `BODY`, `IGNORE`, `MARKDOWN`

## Review Focus Areas

- `TikaGrpcServerImpl.fetchAndParseImpl()`: JSON iteration and `setJsonConfig` calls
- Proto field ordering / backward compatibility (field 5, proto3 — safe)
- e2e test covers HTML and TEXT handler types via `parse_context_json`

## Critical Files

- `tika-grpc/src/main/proto/tika.proto`
- `tika-grpc/src/main/java/org/apache/tika/pipes/grpc/TikaGrpcServerImpl.java`
- `tika-e2e-tests/tika-grpc/src/test/java/org/apache/tika/pipes/filesystem/HandlerTypeTest.java`

## Testing Instructions

Run the tika-grpc e2e tests (requires Docker or a local gRPC server):

```bash
cd tika-e2e-tests/tika-grpc
mvn test -Dtest=HandlerTypeTest
```

## Review Checklist

- [x] Proto change is backward compatible (new field 5, not reusing old number)
- [x] Server handles missing / empty field gracefully (only applies config when non-blank)
- [x] e2e test validates both HTML and TEXT handler configurations
- [x] Unused `BasicContentHandlerFactory` import removed from server impl

## Potential Concerns

- `setJsonConfig` will throw if the component name is not in the registry or the JSON is malformed; the existing gRPC error handling in `fetchAndParseImpl` catches and returns an error status

### DIFF
--- a/tika-grpc/src/main/proto/tika.proto
+++ b/tika-grpc/src/main/proto/tika.proto
@@ -100,6 +100,11 @@ message FetchAndParseRequest {
   string additional_fetch_config_json = 3;
   // The ID of the emitter to use (optional). If not provided, no emitter will be used.
   string emitter_id = 4;
+  // Optional JSON object to configure the ParseContext for this request, overriding server defaults.
+  // Keys are parse-context component names; values are their JSON configs.
+  // Example: {"basic-content-handler-factory": {"type": "HTML"}, "timeout-limits": {"progressTimeoutMillis": 30000}}
+  // See the parse-context.idx component registry for available component names.
+  string parse_context_json = 5;
 }
 
 message FetchAndParseReply {
--- a/tika-e2e-tests/tika-grpc/src/test/java/org/apache/tika/pipes/filesystem/HandlerTypeTest.java
+++ b/tika-e2e-tests/tika-grpc/src/test/java/org/apache/tika/pipes/filesystem/HandlerTypeTest.java
@@ -1,0 +1,356 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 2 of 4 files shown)


tika-grpc/src/main/proto/tika.proto
```diff
@@ -100,6 +100,10 @@ message FetchAndParseRequest {
   string additional_fetch_config_json = 3;
   // The ID of the emitter to use (optional). If not provided, no emitter will be used.
   string emitter_id = 4;
+  // The content handler type to use for this request, overriding the server default.
+  // Valid values: "text" (default), "html", "xml", "body", "ignore", "markdown".
+  // Use "html" to get structured HTML output instead of plain text.
+  string handler_type = 5;
```
nddipiazza: @tballison I think you had intended this to be set via a parse_context_json right? I couldn't figure out how to do it. what do you think?

nddipiazza: nm. claude figured it out.

PR #2803, opened 2026-05-07
Add maxPages option to PDFParserConfig to limit page processing

jnioche:
## Summary                                                                                                                                                                                                                                  
  - Adds `maxPages` field to `PDFParserConfig` (default `-1`, no limit)                                                                                                                                                                     
  - `AbstractPDF2XHTML.processPages()` breaks out of the page loop early when the limit is reached, skipping all text extraction, font mapping, and content stream work for subsequent pages
  - Setter validates that the value is -1 or >= 1, throwing `IllegalArgumentException` otherwise                                                                                                                                              
                                                                                                                                                                                                                                              
  ## Performance                                                                                                                                                                                                                              
  Benchmarked on a 738-page PDF:                                                                                                                                                                                                              
  - Full parse: ~1,800 ms                                                                                                                                                                                                                     
  - First 5 pages: ~135 ms (13× faster)

### DIFF
--- a/tika-parsers/tika-parsers-standard/tika-parsers-standard-integration-tests/src/test/resources/config-examples/pdf-parser-full.json
+++ b/tika-parsers/tika-parsers-standard/tika-parsers-standard-integration-tests/src/test/resources/config-examples/pdf-parser-full.json
@@ -25,6 +25,7 @@
         "imageStrategy": "NONE",
         "maxIncrementalUpdates": 10,
         "maxMainMemoryBytes": 536870912,
+        "maxPages": -1,
         "ocr": {
           "dpi": 300,
           // Options: PNG, TIFF, JPEG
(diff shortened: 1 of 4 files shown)


tballison:
Should we do page range instead? pageStart, pageEnd, with 1 index? That would allow iterating through in chunks.

I'm ok with this as is. wdyt?

jnioche:
> Should we do page range instead? pageStart, pageEnd, with 1 index? That would allow iterating through in chunks.
> 
> I'm ok with this as is. wdyt?

Thanks Tim. Wouldn't that  reload the PDF for each call? or would you treat each chunk as embedded docs?

it would be a more generic approach and still allow  to skip beyond page N - which is what I need for StormCrawler not to waste time getting content that will not be indexed anyway

THausherr:
1 based page range would be in line with what PDFBox text extraction command line does. No it wouldn't reload the document.
However this would be a highly individual setting, and I see tika more of a tool that extracts a mass amount of files.
So I'm neutral about this.

jnioche:
I'll clarify the semantic as there is a confusion between the max number of pages to process vs the largest page number after which the processing should stop. Will also refine the tests shortly
I agree with @THausherr that the range approach would be for specific uses; what I intended to do here is more generic.
@tballison happy to leave it as it is?

tballison:
Thank you, @jnioche (and, of course @THausherr for the review!)

jnioche:
> Thank you, @jnioche (and, of course @THausherr for the review!)

Pleasure! Thanks for reviewing and merging. Great to contribute to Tika again

PR #2807, opened 2026-05-11
TIKA-4725 - updates to docker release workflow

tballison:
make release manual (after the vote)
Change versioning from 4.0.0.0 to 4.0.0-1
Add labels 4.0.0-1 is labeled 4.0.0, and then when 4.0.0-2 comes out we update the label to point to 4.0.0.
Other clean ups we found we needed when pushing 4.0.0-alpha-1.0 from tika-docker.


tballison:
There are multiple ways to go about this. Let me know what you think. This only touches 4.x, and we still have some time to course-correct.

PR #2887, opened 2026-06-09
mask bytes in chm unmarshalUInt32 to avoid sign extension

rootvector2:
`unmarshalUInt32` in `ChmItsfHeader` and `ChmLzxcControlData` ORs the four little-endian bytes from an untrusted chm without masking each `& 0xff`, so any non-leading byte with bit 7 set is sign-extended and corrupts the value (a `0x8000` size/windowSize/resetInterval field reads back as `-32768`); spotted because the sibling readers in `ChmItspHeader`, `ChmPmgiHeader`, `ChmPmglHeader` and `ChmLzxcResetTable` already mask, so this just brings the two stragglers in line.


rootvector2:
any update?

tballison:
I opened a separate ticket to address some of copilot's feedback and to harden the chm parser a bit more: https://issues.apache.org/jira/browse/TIKA-4783 . Thank you for opening this PR @rootvector2 !

PR #2891, opened 2026-06-15
escape link, image and table cell text in ToMarkdownContentHandler

rootvector2:
`escapeMarkdown` is applied to plain text but not to the buffered strings interpolated into link/image/table syntax, so untrusted document content in link text, image `alt`/`src` or a table cell can close the `[..](..)` or `| .. |` context and inject arbitrary markdown; this routes link text, image alt and table cells through `escapeMarkdown`, folds cell newlines that would end a row, and wraps link/image destinations that contain spaces or parens in `<>` where those are allowed.

### DIFF
--- a/tika-core/src/main/java/org/apache/tika/sax/ToMarkdownContentHandler.java
+++ b/tika-core/src/main/java/org/apache/tika/sax/ToMarkdownContentHandler.java
@@ -150,7 +150,8 @@ public class ToMarkdownContentHandler extends DefaultHandler {
             case "img":
                 String alt = atts.getValue("alt");
                 String src = atts.getValue("src");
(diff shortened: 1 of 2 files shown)


tballison:
This is a good catch and LGTM. On a separate issue, I think we should go further, now, and use a markdown library to write the markdown so that we're not whack-a-moling a hand-rolled markdown writer.

rootvector2:
agreed. the buffered link/image/table strings were the only spots that bypassed `escapeMarkdown`, which is how this slipped through. a real renderer on the write side would close the whole class instead of patching each construct one at a time. happy to open a follow-up ticket and take a crack at it if there's appetite.

tballison:
@rootvector2 let me know what you think of: https://github.com/apache/tika/pull/2900

PR #2892, opened 2026-06-15
TIKA-4756: Detect unsigned AcroForm signature fields in PDFs

srujana-kuntumalla:
## Summary

- Adds `PDF.HAS_SIGNATURE_FIELDS` (`pdf:hasSignatureFields`) metadata property to report the presence of AcroForm `/FT /Sig` fields, regardless of whether a signature has been applied
- Refactors `PDFParser.extractSignatures()` to iterate over `PDDocument.getSignatureFields()` instead of `getSignatureDictionaries()`, so unsigned signature fields are detected; `TikaCoreProperties.HAS_SIGNATURE` is still only set when a field has an actual `PDSignature` applied
- Adds a minimal test PDF (`testPDF_unsigned_sig_field.pdf`) with an unsigned signature field and `/SigFlags 3`
- Updates `testSignatureInAcroForm` to assert the new property and tightens expectations; adds `testUnsignedSignatureField` for TIKA-4756

## Motivation

PDFs that contain AcroForm signature fields (`/FT /Sig`) but have not yet been signed were previously indistinguishable from PDFs with no signature infrastructure. This matters for downstream applications such as PDF/A converters (e.g. OCRmyPDF) that need to skip documents with signature fields to avoid invalidating future signatures.

## Test plan

- [x] `PDFParserTest#testUnsignedSignatureField` — new test asserting `pdf:hasSignatureFields=true` and no `HAS_SIGNATURE` on a PDF with unsigned sig field
- [x] `PDFParserTest#testSignatureInAcroForm` — existing test updated to assert the new property; confirms no actual signature is set on `testPDF_acroform3.pdf`

Fixes: https://issues.apache.org/jira/browse/TIKA-4756

### DIFF
--- a/tika-core/src/main/java/org/apache/tika/metadata/PDF.java
+++ b/tika-core/src/main/java/org/apache/tika/metadata/PDF.java
@@ -151,6 +151,13 @@ public interface PDF {
      */
     Property HAS_ACROFORM_FIELDS = Property.internalBoolean(PDF_PREFIX + "hasAcroFormFields");
 
+    /**
+     * Has > 0 AcroForm signature fields (/FT /Sig), regardless of whether a signature
+     * has been applied. Use {@link TikaCoreProperties#HAS_SIGNATURE} to check for
+     * an actual applied digital signature.
+     */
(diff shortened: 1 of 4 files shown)


THausherr:
What's the point of this? Tim has a PR about this (#2889).

PR #2919, opened 2026-06-30
TIKA-4735: Fix --content-only batch output and remove -h/--help short option conflict

nddipiazza:
## Summary

Fixes two bugs in the tika-async-cli `--content-only` batch processing mode.

## Bug 1: `-h` short option conflict with `--help`

The `--help` option in `TikaAsyncCLI` had a `-h` short option alias. This causes option-parsing conflicts.

**Fix:** Removed the `-h` short alias from `--help` in `TikaAsyncCLI`.

## Bug 2: `--content-only` output contained JSON wrapper instead of raw content

When running with `--content-only`, the output files still contained the full JSON metadata wrapper rather than raw text/markdown content.

**Root cause 1 (ParseHandler):** `ParseHandler.parseWithStream()` resolved the effective `ParseMode` (falling back to `PipesConfig.defaultParseMode` when none was set on the request context), but never wrote the resolved mode back into `ParseContext`. `EmitHandler.emit()` reads `ParseContext.get(ParseMode.class)` with no fallback — so it always saw `null` and fell through to the default JSON emit path.

**Fix:** Added `parseContext.set(ParseMode.class, parseMode)` in `ParseHandler.parseWithStream()` immediately after resolving the mode.

**Root cause 2 (EmitHandler passback path):** Even after the ParseHandler fix, the production config template uses `emitStrategy: DYNAMIC` with a 100 KB threshold. For files smaller than 100 KB, `EmitHandler.shouldEmit()` returned `false`, causing the result to be returned as a passback to the parent `AsyncProcessor`. The parent's `AsyncEmitter` then called the batch `emitter.emit(List<EmitData>)` API which serialises metadata as JSON, completely bypassing the `emitContentOnly()` StreamEmitter path.

**Fix:** `EmitHandler.shouldEmit()` now always returns `true` for `CONTENT_ONLY` parse mode (unless the strategy is `PASSBACK_ALL`). This ensures content-only results are always emitted directly from the forked server process via the `StreamEmitter` path, regardless of file size.

### Bug 2 — content-only outputs JSON

Input XML file contains text `main_content` (parent doc) and `some_embedded_content` (embedded doc).

**Before fix:** output file = `[{"X-TIKA:content":"main_content\nsome_embedded_content",...}]`

**After fix:** output file =
```
main\_content

# embed1.xml

some\_embedded\_content
```

The integration test `AsyncProcessorTest#testContentOnlyFromConfigDefault` reproduces this exactly and asserts that the output is raw markdown, not a JSON wrapper.

## Changes

| File | Change |
|------|--------|
| `TikaAsyncCLI.java` | Removed `-h` short option from `--help` |
| `ParseHandler.java` | Write resolved ParseMode back to ParseContext |
| `EmitHandler.java` | CONTENT_ONLY always emits server-side (bypasses DYNAMIC passback) |
| `AsyncProcessorTest.java` | Integration reproducer test |
| `AsyncCliParserTest.java` | Unit tests for CLI arg parsing |
| `TikaConfigAsyncWriterTest.java` | Unit tests for config generation |

## Testing

```
mvn test -pl tika-pipes/tika-async-cli,tika-pipes/tika-pipes-core
```
Results: **65 tests, 0 failures, 0 errors**

### DIFF
--- a/tika-pipes/tika-async-cli/src/main/java/org/apache/tika/async/cli/TikaAsyncCLI.java
+++ b/tika-pipes/tika-async-cli/src/main/java/org/apache/tika/async/cli/TikaAsyncCLI.java
@@ -69,7 +69,7 @@ public class TikaAsyncCLI {
         options.addOption("o", "outputDir", true, "output directory");
         options.addOption("n", "numClients", true, "number of forked clients");
         options.addOption(null, "Xmx", true, "heap for the forked clients, e.g. --Xmx 1g");
-        options.addOption("h", "help", false, "this help message");
+        options.addOption(null, "help", false, "this help message");
         options.addOption("T", "timeoutMs", true, "timeout for each parse in milliseconds");
         options.addOption(null, "handler", true, "handler type: t=text, h=html, x=xml, m=markdown, b=body, i=ignore (default: m)");
         options.addOption("p", "pluginsDir", true, "plugins directory");
--- a/tika-pipes/tika-async-cli/src/test/java/org/apache/tika/async/cli/AsyncCliParserTest.java
+++ b/tika-pipes/tika-async-cli/src/test/java/org/apache/tika/async/cli/AsyncCliParserTest.java
@@ -239,34 +239,4 @@ public class AsyncCliParserTest {
         // Clean up
         Files.deleteIfExists(result);
     }
-
-    // TIKA-4735: --content-only --handler m must correctly parse into SimpleAsyncConfig
-    @Test
-    public void testTIKA4735_ContentOnlyHandlerMarkdown(@TempDir Path tmp) throws Exception {
-        Path inputDir = tmp.resolve("input");
-        Files.createDirectories(inputDir);
-
-        SimpleAsyncConfig config = TikaAsyncCLI.parseCommandLine(
-                new String[]{"--inputDir", inputDir.toString(), "--outputDir", "output",
-                        "--content-only", "--handler", "m"});
-
-        assertTrue(config.isContentOnly(), "TIKA-4735: contentOnly should be true");
-        assertTrue(config.isConcatenate(), "TIKA-4735: --content-only implies --concatenate");
-        assertEquals(BasicContentHandlerFactory.HANDLER_TYPE.MARKDOWN, config.getHandlerType(),
-                "TIKA-4735: handler type should be MARKDOWN");
-    }
-
-    // TIKA-4735: --content-only without --handler should still set contentOnly=true with default MARKDOWN handler
-    @Test
-    public void testTIKA4735_ContentOnlyDefaultHandler(@TempDir Path tmp) throws Exception {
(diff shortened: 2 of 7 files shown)


nddipiazza:
@tballison 
> Bug 1 — -h option conflict (TikaAsyncCLI.java) -h was bound to --help, so using it as a handler shorthand (as documented in the CLI docs) triggered the help menu instead of setting the handler type. Fix: removed the -h short option from --help; --help still works as a long option.

is that ok with you?

nddipiazza:
hmm the test works but the fix does not. still working on it

tballison:
Claude makes these two points. They sound reasonable?

```
Two things worth knowing before it merges

  1. The ParseHandler change is redundant (harmless). The PR re-adds parseContext.set(ParseMode.class, parseMode) for "Root cause 1," but that's already fixed on 
  main/rc1 — PipesWorker.setupParseContext() (commit 0b38268d4, the earlier TIKA-4735 fix) already writes the resolved mode back into the same context object before
  EmitHandler reads it. That's why, on current rc1, you see a filtered [{"X-TIKA:content":"…"}] (filter applied → ParseMode was visible) rather than full metadata.
  So only "Root cause 2" was actually still open. The ParseHandler line is belt-and-suspenders — fine to keep, just not load-bearing.

  2. ⚠️  The tests don't actually lock in this regression. This is the one thing I'd push on. None of the added/changed tests exercise the DYNAMIC-passback path that
  was the bug:
  - AsyncCliParserTest — tests arg parsing only (no emit).
  - TikaConfigAsyncWriterTest — tests generated config has parseMode=CONTENT_ONLY (no emit).
  - The existing AsyncProcessorTest#testContentOnlyFromConfigDefault it cites uses a config with emitStrategy: EMIT_ALL hardcoded — that test passed before this fix
  too, because EMIT_ALL never hits the passback branch.

```

nddipiazza:
Thanks @tballison — addressed both points:

1. **ParseHandler change** — noted it's belt-and-suspenders; kept it since it makes the invariant explicit, but agree it's not the load-bearing fix.

2. **Tests** — removed the 4 tests that didn't exercise the actual changed code (arg-parsing tests in `AsyncCliParserTest`, duplicate config-generation tests in `TikaConfigAsyncWriterTest`). Added `AsyncProcessorTest#testContentOnlyDynamicEmitStrategy` which uses `emitStrategy=DYNAMIC` with a 10 MB threshold so the small test file goes through the passback path — exactly the scenario where the bug manifested. Without the `EmitHandler` fix that test would write JSON; with it, it gets raw text.

tballison:
bq. belt-and-suspenders

Total boss move.

Lol... my claude talking to yours...

tballison:
+1 LGTM

nddipiazza:
yes @tballison  the cute comments it makes sometimes cause me to severely eye roll, but with a smile

PR #2920, opened 2026-07-01
use a per-call DecimalFormat in OggAudioParser duration output

rootvector2:
the static `DURATION_FORMAT` is shared across the ogg parser singletons but `DecimalFormat` is not thread-safe, so concurrent parses race on its mutable internal state; create it per call inside `extractDuration` like `MP4Parser` already does.


hiSandog:
Moving `DecimalFormat` to per-call scope is the right fix for thread safety. One small regression test would make this durable: parse two Ogg inputs concurrently with different duration shapes and assert the emitted metadata is stable across repeated runs. That would catch both shared formatter mutation and any future attempt to reintroduce a static formatter for allocation savings.

rootvector2:
good call. added `OggAudioParserDurationTest` that drives `extractDuration` from 32 threads over a set of rounding-heavy duration shapes and asserts `XMPDM.DURATION` matches the single-threaded value every time. it stays green now and would fail if a shared static formatter ever came back.

rootvector2:
thanks for landing this one.

PR #2935, opened 2026-07-08
TIKA-2861: Extract QuickTime GPS location and item-list metadata in MP4Parser

dschmidt:
JIRA: https://issues.apache.org/jira/browse/TIKA-2861

MP4Parser drops GPS location and other QuickTime metadata from videos:
the box handler skips the `keys`/`ilst` boxes inside `moov/meta`, losing all
`com.apple.quicktime.*` metadata written by modern iOS devices, and the older
`udta` "(c)xyz" path only matched latitude/longitude, dropping the altitude.

Changes (one commit each):

* Parse `keys`/`ilst` in `TikaMp4BoxHandler` and emit UTF-8 text values under
  their key names (e.g. `com.apple.quicktime.content.identifier`). Additive,
  the existing fields are untouched.
* Map the `com.apple.quicktime.location.ISO6709` value to `geo:lat`,
  `geo:long` and `geo:alt` in addition to the raw value.
* Capture the optional ISO 6709 altitude in the "(c)xyz" path and expose it
  as `geo:alt`. Older iOS versions and Google Photos write altitude there;
  Android writes lat/long only, which is unaffected. ExifTool parses the
  same optional component.
* Decode the numeric well-known value types in the item list as well: signed/unsigned
  big endian integers, float32 and float64. Real Live Photo videos carry several of
  those, e.g. `live-photo.vitality-score` (float32) and `live-photo.auto` (uint8),
  which were silently dropped before.

Alternatives considered:

* Switching MP4Parser to a different reader (e.g. metadata-extractor's
  QuickTime/MOV reader, which does handle this metadata): rejected as much
  higher risk, since it would change the parser's existing output wholesale
  instead of adding the missing fields.
* Fixing this upstream in metadata-extractor's `Mp4BoxHandler`: awkward,
  since it would duplicate functionality of their QuickTime reader, and
  upstream would likely (and reasonably) answer that MOV files should just
  be read with that reader instead. Tika already extends the box handler
  with Tika-specific boxes, so this follows the established pattern.

Testing: two new minimal fixtures (`testMP4_QuickTimeMetadata.mov` for the
keys/ilst path, `testMP4_udtaLocation.mp4` for the previously untested udta
path) with `MP4ParserTest` assertions for the raw values and geo properties.


dschmidt:
Tested with a real Live Photo pair from the osxphotos test libraries (iPhone 15 Pro, iOS 18.5):

https://github.com/RhetTbull/osxphotos/blob/main/tests/Test-Live-15.7.2.photoslibrary/originals/C/C3090F66-942C-41D3-BEC7-4F2B4876A109.heic
https://github.com/RhetTbull/osxphotos/blob/main/tests/Test-Live-15.7.2.photoslibrary/originals/C/C3090F66-942C-41D3-BEC7-4F2B4876A109_3.mov

With a server built from this branch, `/rmeta` returns the same content identifier for both files now (`CD28D161-D5EC-4CDE-8B60-DACCC1363B6B`, as `img:Content Identifier` on the heic and `com.apple.quicktime.content.identifier` on the mov), so matching the two halves of a Live Photo works. Before, the mov side had no identifier at all. The location shows up as `geo:lat`/`geo:long`/`geo:alt` (`+36.0970-081.7844+1330.402/`).

Also checked the .MOV attached to the JIRA issue, its location `-35.2844+149.1214+573.764/` comes through including the altitude.

dschmidt:
When comparing output for live photos I've noticed that some output was missing and that I should have not been handling only strings 🙃

tballison:
Sorry, should have confirmed license before merging. The two photos added were from the link above? 

If so, looks like MIT so we're good?

https://github.com/RhetTbull/osxphotos/blob/main/LICENSE

Otherwise, you can confirm they are ASL 2.0?

@dschmidt

dschmidt:
No license concern there: the two committed test files are not photos at
all. Both are fully synthetic, hand-crafted container structures
(`testMP4_QuickTimeMetadata.mov` is 611 bytes, `testMP4_udtaLocation.mp4`
is 74 bytes): just an `ftyp` box plus a `moov` with the metadata boxes
under test, no `mdat`, no image or video data, and all values are made up
(`TEST-UUID-0001-LIVEPHOTO`, `+12.3456-098.7654+010.500/`).

The osxphotos files linked above (MIT) were only used locally for the
end-to-end verification described in my earlier comment; they were never
added to the repo.

THausherr:
cherry-picked for 3.0

PR #2938, opened 2026-07-09
Follow on tweaks to the recent TIKA-2861 fix

tballison:
Simplify regex Pattern. Follow up to https://github.com/apache/tika/pull/2935

### DIFF
--- a/tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp4/TikaMp4BoxHandler.java
+++ b/tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp4/TikaMp4BoxHandler.java
@@ -23,8 +23,6 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import com.drew.imaging.mp4.Mp4Handler;
 import com.drew.lang.annotations.NotNull;
@@ -35,6 +33,7 @@ import com.drew.metadata.mp4.Mp4Context;
 import org.xml.sax.SAXException;
 
 import org.apache.tika.metadata.TikaCoreProperties;
+import org.apache.tika.parser.mp4.boxes.ISO6709;
 import org.apache.tika.parser.mp4.boxes.TikaUserDataBox;
 import org.apache.tika.sax.XHTMLContentHandler;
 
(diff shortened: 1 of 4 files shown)


dschmidt (commented):
Sweet, thanks for the cleanup!

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp4/TikaMp4BoxHandler.java
```diff
@@ -186,12 +183,12 @@ private void processQuickTimeItemList(@Nullable byte[] payload) {
...
-            tikaMetadata.set(TikaCoreProperties.LONGITUDE, Double.parseDouble(matcher.group(2)));
-            if (matcher.group(3) != null) {
-                tikaMetadata.set(TikaCoreProperties.ALTITUDE, Double.parseDouble(matcher.group(3)));
+        ISO6709.Location location = ISO6709.parse(iso6709);
+        if (location != null) {
+            tikaMetadata.set(TikaCoreProperties.LATITUDE, location.latitude);
+            tikaMetadata.set(TikaCoreProperties.LONGITUDE, location.longitude);
+            if (location.altitude != null) {
+                tikaMetadata.set(TikaCoreProperties.ALTITUDE, location.altitude);
             }
```
dschmidt, line 192: Wondering if tikaMetadata could be made aware of the ISO6709.Location type? Then 5 lines would shrink to one ... but maybe it's not a good idea for coupling reasons 🤷🏻‍♂️

dschmidt: FWIW it might come in handy working on this https://issues.apache.org/jira/browse/TIKA-4781?

dschmidt: Oh, we're not dealing with ISO6709 there - nvm :)

dschmidt: ... unless it would become a generic Location class - with `parseFromIso6709()`, `parseSomething()` methods that would also parse other formats as well?

Was never a Java person, so not sure what's really idiomatic :)

tballison: Y, not clear for coupling reasons? Definitely a separate issue and ticket. As long as altitude is optional, I think this makes a bunch of sense. wdyt? This would allow us to have arrays of points rather than parallel arrays, which would be so much better.

PR #2943, opened 2026-07-10
TIKA-4784 -- improve utf16 handling in id3 

tballison:

* the issue ID (`TIKA-XXXX`)

### DIFF
--- a/tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp3/ID3v22Handler.java
+++ b/tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp3/ID3v22Handler.java
@@ -72,7 +72,7 @@ public class ID3v22Handler implements ID3Tags {
                     copyright = getTagString(tag.data, 0, tag.data.length);
                     break;
                 case "COM":
-                    comments.add(getComment(tag.data, 0, tag.data.length));
+                    addComment(getComment(tag.data, 0, tag.data.length));
                     break;
                 case "TRK":
                     trackNumber = getTagString(tag.data, 0, tag.data.length);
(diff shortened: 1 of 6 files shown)


tballison:
cc @dschmidt wdyt?

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp3/ID3v22Handler.java
```diff
@@ -113,6 +113,13 @@ private ID3Comment getComment(byte[] data, int offset, int length) {
         return ID3v2Frame.getComment(data, offset, length);
     }
 
+    /** Skips null comments (malformed frames) that would trip up consumers. */
+    private void addComment(ID3Comment comment) {
+        if (comment != null) {
+            comments.add(comment);
+        }
+    }
```
dschmidt: major nitpick: tbh when I saw a helper function being introduced, I expected it to do more than j́ust a simple null check. Same as underneath. I would probably do it inline if the function isnt called a bunch of times

dschmidt (commented):
LGTM. I'm trying to avoid charset handling wherever I can so my knowledge is a bit limited on the topic, but from what I and my agent can tell it's reasonable change.

I'm so happy tika deals with this kind of stuff, so my downstream applications don't have to :)

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp3/ID3v22Handler.java
tballison: Lol, y, that's bad.

dschmidt: because I expect a function like addComment to do more - but in the end: I'm obviously not emotional about that

PR #2947, opened 2026-07-11
TIKA-4779: Add audio:track-count and audio:disc-count, normalize n/total values

dschmidt:
JIRA: https://issues.apache.org/jira/browse/TIKA-4779

The total number of tracks and discs is present in all common audio containers but was dropped: the MP4 parser read the second value of the `trkn`/`disk` atoms and discarded it, ID3 TRCK/TPOS "n/total" values were passed through unsplit, and the Vorbis TRACKTOTAL/DISCTOTAL comments were not mapped.

XMPDM has no property for the totals and cannot be extended, so this adds an `org.apache.tika.metadata.Audio` interface under Tika's own namespace, following the pattern of `Geographic` and the Google properties from TIKA-4775:

* `audio:track-count` / `audio:disc-count`: the totals, from the MP4 atoms, the "n/total" form, and the Vorbis TRACKTOTAL/TOTALTRACKS/DISCTOTAL/TOTALDISCS comments (explicit comments win over the combined form).
* `audio:raw-track-number` / `audio:raw-disc-number`: the value exactly as tagged.

With the raw value preserved, `xmpDM:trackNumber` and `xmpDM:discNumber` now only receive clean positive integers. They are declared as Integer properties but previously received the raw tag value, including "3/12" and non-numeric forms like vinyl "A1", which broke typed consumers. This is a deliberate output change that should land before 4.0; nothing is lost, the original value stays available under the raw properties. The combined form is parsed by a small `NumberAndTotal` value class, same shape as the recently added `ISO6709` parser. The Ogg path also gains the previously unmapped DISCNUMBER comment.

This is part 1 of 2 for TIKA-4779; bitrate, VBR detection and DRM detection need actual parser work (Xing/VBRI headers, MP4 `sinf`) and will come separately.

Testing: the existing `testMP4.m4a` fixture already carries the totals (track 1 of 42, disc 6 of 12, previously discarded); the ID3v2.4 fixture gains a TRCK "3/12" frame (25 bytes appended to the tag, audio bytes untouched); `NumberAndTotalTest` covers the parse edge cases and the parser tests assert typed, count and raw values including degenerate forms.

One process question: this PR deliberately changes metadata output shape (xmpDM:trackNumber/discNumber becoming strictly integer), and 4.0 is accumulating a few more of those (e.g. the xmp-raw rename in TIKA-4775). Should such output changes be recorded in the BREAKING CHANGES section of CHANGES.txt as a matter of course? Happy to add an entry here, I just did not want to unilaterally open a new unreleased section in the file.


tballison:
Agent has one small nit: 
```
TikaUserDataBox.java:164-167 — the disk branch reads a fixed 6 bytes with no length guard, unlike the adjacent
  trkn branch (PLAUSIBLE, pre-existing but in the touched branch). The trkn case guards if (toRead == 8) before
  reading 8 bytes and skips otherwise; the disk case unconditionally reads getInt32()+getInt16() (6 bytes) then
  does totalRead += toRead. If any encoder writes a disk atom whose data length ≠ 6 (e.g. an 8-byte form mirroring
  trkn), totalRead diverges from the bytes actually consumed and the remaining ilst records are read misaligned.
  Real-world files are 6 bytes so this is latent, but since the PR now reads b from this branch it's the natural
  spot to add the same toRead ==-style guard/skip the trkn branch already has.
```

dschmidt:
Good catch, fixed in 5a69b1568. The disk branch now reads the six known
bytes when at least that much is declared and skips the remainder, so the
walk always consumes exactly the declared length (padded 8-byte disk
atoms now also yield the disc total instead of derailing the parse).

To make sure the fix bites, the udta fixture gained an iTunes-style
meta/mdir/ilst with a padded 8-byte disk atom followed by a title entry;
without the guard the misaligned walk takes down the rest of the parse,
which the test now pins.

dschmidt:
Amazing, thanks! This has been bugging me for so incredibly long but I never found the motivation to work on it ... Thanks, Claude, I guess :D

PR #2965, opened 2026-07-23
TIKA-4794-option-2: tk normalization

tballison:
This would be the first proposed change (option 2) here: https://issues.apache.org/jira/browse/TIKA-4794


tballison:
Absent feedback, we're implementing this and #2964 . #2964 build off of this and includes everything here. No need for a separate PR.

PR #2968, opened 2026-07-28
TIKA-4796: compile MagicDetector regex once at construction

gsingers:
## Summary
- `MagicDetector` recompiled its `Pattern` on every regex match even though the pattern bytes and case-insensitivity flag are fixed at construction time; now compiled once in the constructor and reused (`Pattern` is immutable, `Matcher` is still created per call, so the regex path stays thread-safe).
- Adds direct coverage for the regex branch of `matches(byte[])`, which `MagicMatch.eval` exercises for every magic in `tika-mimetypes.xml` but which was previously only tested indirectly, plus repeated-call and concurrent-use tests over a shared detector instance.
- CHANGES.txt entry added.

## Test plan
- [x] `./mvnw clean test -pl :tika-core -Dtest=MagicDetectorTest`
- [x] `./mvnw clean install`

https://claude.ai/code/session_01Pp6TVDKF7ztw7SS7hwwuFd

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -6,6 +6,11 @@ Release 4.0.0 - ???
      opt-in Pkcs7Detector surfaces the subtype at detect() time,
      but must be enabled via configuration (TIKA-1997).
 
+  OTHER CHANGES
+
+   * MagicDetector now compiles its regular expression once, in the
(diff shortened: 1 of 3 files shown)


gsingers:
I will open a backport branch for this as well.

PR #2972, opened 2026-07-28
[TIKA-4796] compile MagicDetector regex once at construction (3.x backport)

gsingers:
Backport of #2968 (`65a2f3ab6`) from `main` to `branch_3x`.

JIRA: https://issues.apache.org/jira/browse/TIKA-4796

## What

`MagicDetector` recompiled its `Pattern` on every regex `detect()` call, even though both inputs to the compile — the pattern bytes and the case-insensitivity flag — are fixed at construction time. This compiles it once in the constructor and reuses it.

`Pattern` is immutable and a new `Matcher` is still created per call, so the regex path remains safe for the shared, long-lived detector instances that `MimeTypes` hands out.

## Not a cherry-pick

3.x compiles the `Pattern` inline in `detect()`; `main` had already extracted the matching logic into `matchesBuffer()` behind `matches(byte[])`. The change is therefore hand-ported to the 3.x structure rather than cherry-picked.

The same applies to the tests. The `main` PR added `testMatchesByteArrayRegEx` to cover the regex branch of `matches(byte[])`, which does not exist on 3.x — and 3.x already exercises the regex `detect()` path through `testDetectRegExPDF`, `testDetectRegExGreedy` and `testDetectRegExOptions`. Only the two guards that apply here are carried over, rewritten against `detect(InputStream, Metadata)`:

- `testRegExDetectorRepeatedCallsStable` — repeated calls on one instance stay independent
- `testRegExDetectorConcurrent` — 8 threads x 200 iterations against a shared detector

## Behavior change

A malformed regex now raises `PatternSyntaxException` from the constructor rather than from the first `detect()`. Via `MagicMatch` this is still lazy, since `getDetector()` builds on first `eval()`, so loading `tika-mimetypes.xml` is unaffected. Documented with `@throws` on the constructor.

## Testing

`./mvnw clean install -pl :tika-core` — 416 tests, 0 failures; `MagicDetectorTest` 13/13. Checkstyle and Spotless clean.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,3 +1,8 @@
+Release 3.3.3 - ???
+
+  * MagicDetector now compiles its regular expression once, in the
+    constructor, instead of recompiling it on every match (TIKA-4796).
+
 Release 3.3.2 - 7/16/2026
 
   * Dependency upgrades (TIKA-4738).
--- a/tika-core/src/main/java/org/apache/tika/detect/MagicDetector.java
+++ b/tika-core/src/main/java/org/apache/tika/detect/MagicDetector.java
@@ -39,6 +39,9 @@ import org.apache.tika.mime.MediaType;
  * Because this works on bytes, not characters, by default any string
  * matching is done as ISO_8859_1. To use an explicit different
  * encoding, supply a type other than "string" / "stringignorecase"
+ * <p>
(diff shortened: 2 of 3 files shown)


tballison:
Thank you @gsingers !

PR #2981, opened 2026-07-30
TIKA-4776: Add regression test for Apple maker note LONG8 tags

dschmidt:
With metadata-extractor 2.21.0 on main the Apple maker note tags written as LONG8 (TIFF format code 16) come through, so this adds the regression test discussed in the ticket. The fixture is the still half of a real Apple Live Photo (iPhone 15 Pro, iOS 18.5) from the MIT licensed osxphotos test suite; the test asserts the content identifier and the Live Photo ID (maker note tag 0x0017, value 5283876), which metadata-extractor 2.20.0 silently dropped. With this in place TIKA-4776 can be resolved.


THausherr:
Would it be possible to create a smaller file? Yours has a 1 MB size.

dschmidt:
Good point. The fixture is now 3.2 KB instead of 1.4 MB: the real EXIF payload (including the Apple maker note with the LONG8 tag, byte for byte from the original photo) repackaged into a minimal HEIC container, so the regression still exercises the real-world bytes without shipping the full image.

dschmidt:
The CI failure is unrelated to the change: both jobs died within a minute while Maven resolved the nisse build extension from Central (transient repository hiccup, nothing was compiled yet). Could someone with the permissions restart the failed jobs?

tballison:
Mea culpa. I broke main. Merge conflict without warning. Fixing now.

https://github.com/apache/tika/pull/2985

dschmidt:
Thanks :)

tballison:
Main should be working again. Please merge and see what happens.

dschmidt:
done :)

dschmidt:
CI is green :)

PR #2982, opened 2026-07-30
TIKA-4800: Expose the MP4 frame rate, audio channels and sample size

dschmidt:
JIRA: https://issues.apache.org/jira/browse/TIKA-4800

metadata-extractor's `Mp4VideoDirectory` and `Mp4SoundDirectory` already populate the frame rate, channel count and audio sample size, but `MP4Parser` only mapped a subset. This forwards the three: `video:frame-rate` (a new `Video` property, since `XMPDM.VIDEO_FRAME_RATE` is a closed-choice text field), `audio:channels` (the raw count; `XMPDM.AUDIO_CHANNEL_TYPE` only distinguishes Mono/Stereo) and `audio:bits-per-sample`.

One property per commit. Tests assert channels/bits against the existing `testMP4.m4a` (2 channels, 16 bits) and the frame rate against a new 1.8 KB ffmpeg-generated clip (`testMP4Video.mp4`, 10 fps).


tballison:
@dschmidt able to resolve conflicts? Thank you!

dschmidt:
Yes again 🤓

PR #2983, opened 2026-07-30
TIKA-4801: Extract cover art from MP3, MP4, Vorbis and FLAC audio

dschmidt:
Fixes TIKA-4801: ID3v2 APIC frames (v2.2 PIC included), the MP4/M4A `covr` atom (previously explicitly skipped), Vorbis/Opus `metadata_block_picture` comments and native FLAC PICTURE blocks.

Each picture is emitted through the standard EmbeddedDocumentExtractor mechanism as an inline embedded document, the same way PDF and Office parsers emit embedded images: /rmeta returns one metadata entry per cover with the declared content type (or auto-detection where MP4 only has a type code), the declared description as dc:title and the picture type as dc:description; /unpack delivers the bytes. Deliberately no new metadata properties on the audio document and no mapping of the width/height fields that Vorbis picture blocks declare: ID3 and MP4 declare no dimensions at all, and the embedded entry carries the real ones once image parsers are on the classpath.

The raw metadata_block_picture comment is no longer passed through as a giant base64 vorbis: value, it becomes the embedded document instead.

Native FLAC PICTURE blocks are read by a small metadata block walk in FlacParser itself: vorbis-java 0.8 offers no public accessor for the blocks (their master already has FlacFile.getOtherMetadata(), but nothing has been released since 0.8), so the parser walks the native metadata region through a second stream and reuses the shared picture handler. A TODO in the code links Gagravarr/VorbisJava#46 so the walk can be dropped once a release ships. FLAC-in-Ogg needs no extra path, pictures arrive there as metadata_block_picture.

One commit per format, each with its own fixture and end-to-end test; the fixtures are minimally modified copies of existing test files carrying a 1x1 PNG cover (for the Ogg fixture the comment packet was rewritten with recomputed page CRC). All four tests fail without their parser change.


tballison:
This is amazing. All the nooks and crannies for embedded files...

dschmidt:
Sweeeeeeeeet 🤩

any chance for a new beta anytime soon?
Also what's the plan when to release 4.x final/stable?

PR #3003, opened 2026-08-10
TIKA-4812 - improve media file robustness

tballison:

* the issue ID (`TIKA-XXXX`)

### DIFF
--- a/tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp3/ID3v2Frame.java
+++ b/tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp3/ID3v2Frame.java
@@ -670,8 +670,12 @@ public class ID3v2Frame implements MP3Frame {
         }
 
         public boolean hasNext() {
-            // Check for padding at the end
-            return offset < data.length && data[offset] != 0;
+            // Stop at padding, and at a truncated tail too short for a full frame
+            // header: the RawTag constructor reads the header bytes unconditionally,
+            // so without the data.length no longer being zero-padded (TIKA-4812) a
+            // partial header would throw ArrayIndexOutOfBoundsException.
+            return offset + nameLength + sizeLength + flagLength <= data.length
+                    && data[offset] != 0;
         }
 
         public RawTag next() {
--- a/tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp3/ID3v2Frame.java
+++ b/tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp3/ID3v2Frame.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.PushbackInputStream;
 import java.io.UnsupportedEncodingException;
+import java.util.Arrays;
 import java.util.Iterator;
 
 import org.apache.tika.parser.mp3.ID3Tags.ID3Comment;
@@ -180,9 +181,10 @@ public class ID3v2Frame implements MP3Frame {
                     throw new IOException("Tried to read " + length + " bytes, but only " + pos +
                             " bytes present");
                 } else {
-                    // Give them what we found
-                    // TODO Log the short read
-                    return b;
+                    // truncated stream: return only the bytes actually read, not the
+                    // zero-padded full-length array, so callers (e.g. cover-art
+                    // extraction) don't emit padding as data. TIKA-4812
+                    return Arrays.copyOf(b, pos);
                 }
             }
             pos += read;
--- a/tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp4/TikaMp4Reader.java
+++ b/tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp4/TikaMp4Reader.java
@@ -43,53 +43,76 @@ final class TikaMp4Reader {
     private TikaMp4Reader() {
     }
 
-    static void extract(InputStream inputStream, Mp4BoxHandler handler, long maxBoxSize) {
+    //MP4 containers nest (moov/trak/mdia/minf/stbl/udta/meta); cap the recursion so a
+    //crafted chain of nested container headers cannot overflow the stack (an uncaught
+    //Error, caught by neither the IOException handler below nor CompositeParser). Real
+    //files nest well under this.
+    private static final int MAX_BOX_DEPTH = 100;
+
+    /**
+     * @param inputLength total input length in bytes, or -1 if unknown. When known, a box
+     *                    that declares more payload than the input holds is skipped rather
+     *                    than allocated (StreamReader.getBytes allocates before reading).
+     */
+    static void extract(InputStream inputStream, Mp4BoxHandler handler, long maxBoxSize,
+                        long inputLength) {
         StreamReader reader = new StreamReader(inputStream);
         reader.setMotorolaByteOrder(true);
-        processBoxes(reader, -1, handler, new Mp4Context(), maxBoxSize);
+        processBoxes(reader, -1, handler, new Mp4Context(), maxBoxSize, inputLength, 0);
     }
 
     private static void processBoxes(StreamReader reader, long atomEnd, Mp4Handler<?> handler,
-                                     Mp4Context context, long maxBoxSize) {
+                                     Mp4Context context, long maxBoxSize, long inputLength,
+                                     int depth) {
+        if (depth > MAX_BOX_DEPTH) {
+            handler.addError("MP4 box nesting exceeds the maximum depth of " + MAX_BOX_DEPTH);
+            return;
+        }
         try {
             while (atomEnd == -1 || reader.getPosition() < atomEnd) {
                 long boxSize = reader.getUInt32();
                 String boxType = reader.getString(4);
-                boolean isLargeSize = boxSize == 1;
-                if (isLargeSize) {
+                //4 bytes size + 4 bytes type, plus 8 more when a 64-bit largesize follows
+                int headerSize = boxSize == 1 ? 16 : 8;
+                if (headerSize == 16) {
                     boxSize = reader.getInt64();
                 }
                 if (boxSize > Integer.MAX_VALUE) {
                     handler.addError("Box size too large.");
                     break;
                 }
-                if (boxSize < 8) {
+                if (boxSize < headerSize) {
                     handler.addError("Box size too small.");
                     break;
                 }
 
+                long payloadLength = boxSize - headerSize;
                 if (acceptContainer(handler, boxType)) {
-                    processBoxes(reader, boxSize + reader.getPosition() - 8,
+                    processBoxes(reader, reader.getPosition() + payloadLength,
                             processBox(handler, boxType, null, boxSize, context), context,
-                            maxBoxSize);
+                            maxBoxSize, inputLength, depth + 1);
                 } else if (acceptBox(handler, boxType)) {
-                    long payloadLength = boxSize - 8;
-                    if (payloadLength > maxBoxSize) {
+                    //StreamReader.getBytes allocates the whole payload up front, so skip
+                    //(a lazy stream advance) any box over the cap, or one that claims more
+                    //than the input holds, instead of allocating it. Skip-and-continue is
+                    //deliberate: unlike the TikaMemoryLimitException other parsers throw,
+                    //this keeps the remaining boxes' metadata; the skip is recorded as a
+                    //warning via the directory's error list.
+                    boolean tooLarge = payloadLength > maxBoxSize;
+                    boolean beyondInput = inputLength >= 0
+                            && reader.getPosition() + payloadLength > inputLength;
+                    if (tooLarge || beyondInput) {
                         handler.addError("MP4 box '" + boxType + "' payload (" + payloadLength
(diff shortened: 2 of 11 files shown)


tballison:
@dschmidt let me know what you think of these. These are underlying issues that predate your recent PRs.

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp3/ID3v2Frame.java
```diff
@@ -180,9 +181,10 @@ protected static byte[] readFully(InputStream inp, int length, boolean shortData
                     throw new IOException("Tried to read " + length + " bytes, but only " + pos +
                             " bytes present");
                 } else {
-                    // Give them what we found
-                    // TODO Log the short read
-                    return b;
+                    // truncated stream: return only the bytes actually read, not the
+                    // zero-padded full-length array, so callers (e.g. cover-art
+                    // extraction) don't emit padding as data. TIKA-4812
+                    return Arrays.copyOf(b, pos);
```
dschmidt, line 187: Claude says:
```
The zero-padding this removes was load-bearing for the ID3v2 tag body read, which uses
shortDataIsFatal=false (line 68). RawTagIterator.hasNext() only checks
offset < data.length && data[offset] != 0, and the RawTag constructor then reads the
name, size, and flag bytes unconditionally. With the array shortened, an MP3 whose tag
is cut off mid-frame-header (e.g. an interrupted download ending with "TI" of "TIT2")
now throws ArrayIndexOutOfBoundsException out of the ID3v2x handler constructors, which
Mp3Parser.parse does not catch. Previously such files parsed cleanly with partial
metadata. Suggestion: keep the truncated copy, but make RawTagIterator stop iterating
when fewer than a full frame header's bytes remain (or bounds-check the header reads).
```

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp4/TikaMp4Reader.java
```diff
@@ -0,0 +1,126 @@
...
+                if (boxSize > Integer.MAX_VALUE) {
+                    handler.addError("Box size too large.");
+                    break;
+                }
+                if (boxSize < 8) {
+                    handler.addError("Box size too small.");
+                    break;
+                }
+
+                if (acceptContainer(handler, boxType)) {
```
dschmidt, line 90: Claude says:
```
processBoxes recurses once per accepted container with no depth cap, which is the same
StackOverflowError bug class this PR fixes in TikaMp4SoundHandler and FLVParser.
Mp4BoxHandler.shouldAcceptContainer accepts moov/trak/mdia/udta/meta by fourCC alone,
so a crafted file that is a chain of nested "moov" headers (8 bytes per level) drives
one stack frame per level; a few hundred KB of input overflows the default JVM stack.
StackOverflowError is an Error, so neither the IOException catch here nor
CompositeParser will contain it. Since Tika now owns this walk, threading a depth
counter through processBoxes and doing addError+break past a small limit would close it.
```

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp4/TikaMp4Reader.java
```diff
@@ -0,0 +1,126 @@
...
+                    processBoxes(reader, boxSize + reader.getPosition() - 8,
+                            processBox(handler, boxType, null, boxSize, context), context,
+                            maxBoxSize);
+                } else if (acceptBox(handler, boxType)) {
+                    long payloadLength = boxSize - 8;
```
dschmidt: Claude says:
```
For 64-bit largesize boxes (boxSize == 1), 16 header bytes have been consumed, but the
accepted-box path computes payloadLength = boxSize - 8 and the container path computes
atomEnd = boxSize + position - 8; both overrun into the next sibling box by 8 bytes.
Only the unaccepted-box branch uses the correct boxSize - 16. This faithfully mirrors a
metadata-extractor bug, but now that the walk is Tika-owned it seems worth fixing:
track headerSize (8 or 16) and use boxSize - headerSize in all three places. As is, a
legal file using largesize encoding gets its remaining boxes misparsed ("Box size too
small/large") and silently loses metadata from that point on.
```

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp4/TikaMp4Reader.java
```diff
@@ -0,0 +1,126 @@
...
+                    if (payloadLength > maxBoxSize) {
+                        handler.addError("MP4 box '" + boxType + "' payload (" + payloadLength
```
dschmidt, line 105: Claude says:
```
Minor consistency point: other parsers (BPGParser, PSDParser, ICNSParser,
AppleSingleFileParser) throw TikaMemoryLimitException when a declared length exceeds a
configured max, while this branch records a string in the Mp4Directory error collection
(itself capped by MAX_ERROR_MESSAGES) and continues. Skip-and-continue may well be the
right call for robustness here, but if so it might deserve a comment documenting the
deliberate deviation; otherwise throwing TikaMemoryLimitException would match the
convention and make truncated extraction observable to operators.
```

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp4/TikaMp4Reader.java
```diff
@@ -0,0 +1,126 @@
...
+                                + " bytes) exceeds the maximum of " + maxBoxSize
+                                + " bytes; skipping.");
+                        reader.skip(payloadLength);
+                    } else {
+                        handler = processBox(handler, boxType,
+                                reader.getBytes((int) payloadLength), boxSize, context);
```
dschmidt, line 112: Claude says:
```
StreamReader.getBytes allocates the full buffer before reading any data, so an 8-byte
crafted header still forces a maxBoxSize-sized allocation (100MB by default) from a
near-empty stream, and a batch of such tiny files creates sustained GC pressure in
tika-server. Legitimate metadata boxes are well under 1MB. Options: clamp payloadLength
against the bytes plausibly available before allocating, or default the cap much lower
(a few MB) since it is configurable.
```

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp4/boxes/TikaUserDataBox.java
```diff
@@ -123,6 +123,12 @@ private void parseUserDataBox(SequentialReader reader, String handlerType,
         //this handles "free" types...not sure if there are others?
         //will throw IOException if no ilist is found
         while (! subType.equals(ILST)) {
+            //re-validate each re-read length: len < 8 makes skip(len - 8) negative,
+            //which throws IllegalArgumentException (not IOException, so it escapes
+            //MP4Reader). See TIKA-4812.
+            if (len < 8L || len >= Integer.MAX_VALUE) {
```
dschmidt, line 126: Claude says:
```
Two notes on this guard. First, the len >= Integer.MAX_VALUE half silently returns
where the old skip-to-EOF path threw an EOFException that Mp4Reader caught and recorded
as a parse error; the outer while (reader.getPosition() < length) loop in the
constructor then resumes at an unaligned offset inside the meta payload and can
misparse garbage as box headers (worst case emitting bogus geo coordinates via the
"(c)xyz" branch). Recording an error and/or aborting the udta walk would preserve the
old observability. Second, this near-duplicates the pre-loop check at line 118
(len >= Integer.MAX_VALUE || len <= 0) with a different lower bound, so len in [1,7]
passes one guard and fails the other. Restructuring so each length is validated exactly
once right after it is read would keep the two from drifting apart.
```

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp4/boxes/TikaUserDataBox.java
```diff
@@ -185,6 +191,11 @@ private void processIList(SequentialReader reader, long totalLen)
                 } else if ("cpil".equals(fieldName)) {
                     int compilationId = (int)reader.getByte();
                     metadata.set(XMPDM.COMPILATION, compilationId);
+                    //consume the rest of the declared field: totalRead counts toRead,
+                    //but only 1 byte was read, so skip the remainder to stay aligned
+                    if (toRead > 1) {
```
dschmidt: Claude says:
```
This cpil fix is correct, but the underlying issue is general: processIList aligns
per-branch from fieldLen instead of realigning to recordStart + recordLen after each
record (pre-existing, same bug class this fix addresses). Any ilst record whose data
atom is followed by a trailing sub-atom (e.g. a 12-byte "name" atom, which
iTunes-style writers can emit) leaves bytes unconsumed; the next iteration misreads
them as a record header and the toSkip <= 0 path silently drops all remaining ilst
metadata. Recording the position at the loop top and skipping to recordStart +
recordLen after each branch would make every branch alignment-safe at once.
```

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/video/FLVParser.java
```diff
@@ -124,11 +135,11 @@ private Object readAMFData(DataInputStream input, int type) throws IOException {
         }
     }
 
-    private Object readAMFStrictArray(DataInputStream input) throws IOException {
+    private Object readAMFStrictArray(DataInputStream input, int depth) throws IOException {
```
dschmidt, line 144: Claude says:
```
The depth bound fixes the nesting case, but these loops remain width-unbounded:
readAMFStrictArray/readAMFEcmaArray take count/size from readUInt32 (up to ~4 billion)
and grow an ArrayList/HashMap per iteration with no validation against remaining input.
A single flat array with a huge declared count backed by cheap repeated bytes runs to
OutOfMemoryError without ever hitting MAX_AMF_DEPTH. Capping count against a
max-element constant (mirroring the depth bound) would close the width half too.
```

tika-parsers/tika-parsers-standard/tika-parsers-standard-modules/tika-parser-audiovideo-module/src/main/java/org/apache/tika/parser/mp4/TikaMp4SoundHandler.java
```diff
@@ -103,12 +103,25 @@ private static int soundEntrySize(int version) {
...
+    //stack overflows (an uncaught Error, not caught by Mp4Reader or CompositeParser).
+    //See TIKA-4812.
+    private static final int MAX_BOX_DEPTH = 10;
+
     /**
      * Scans the child boxes of a sample entry for an 'esds' box and returns
      * its average bitrate, or 0 if there is none. QuickTime version 1/2
      * entries may nest the 'esds' inside a 'wave' extension box.
      */
     private static int findEsdsAverageBitRate(byte[] b, int pos, int end) {
```
dschmidt: Claude says:
```
Tiny nit: this 3-arg wrapper exists only to pass depth 0 and has exactly one caller
(unlike FLVParser's pair, where a test needs the short signature). Passing 0 at the
single call site and deleting the wrapper would keep the javadoc and the implementation
on the same method.
```

dschmidt (commented):
Claude says:
> Careful review of this PR. All five hardening changes address real issues, and the cpil skip and the FLV depth threading check out as written. Three points I would consider blocking, detailed inline:
>
> 1. **ID3v2Frame truncation change is a regression on real-world truncated MP3s.** The removed zero-padding was load-bearing for RawTagIterator/RawTag, which read frame headers without bounds checks; files that previously parsed with partial metadata now throw an unchecked ArrayIndexOutOfBoundsException.
> 2. **TikaMp4Reader itself has no recursion depth cap.** Nested "moov" containers drive one stack frame per level, the exact StackOverflowError bug class this PR fixes elsewhere, and nothing catches the Error.
> 3. **The largesize (boxSize == 1) paths are off by 8 bytes.** payloadLength and atomEnd use boxSize - 8 although 16 header bytes were consumed; this copies a metadata-extractor bug into now-Tika-owned code and silently breaks legal files using largesize encoding.
>
> Non-blocking, also inline: the allocate-before-read in StreamReader.getBytes keeps a 100MB-per-tiny-file allocation possible under the default cap; the AMF fix bounds depth but not array width; the new TikaUserDataBox guard silently swallows what used to be a recorded parse error and near-duplicates the pre-loop check with a different lower bound; processIList's per-branch alignment (the general form of the cpil bug) is still fragile; the over-limit skip diverges from the TikaMemoryLimitException convention; and the findEsdsAverageBitRate wrapper could be inlined.
>
> Checked and deliberately not raised: the maxBoxSize setter binds fine via Tika 4 JSON config without @Field, and the instanceof handler dispatch in TikaMp4Reader is safe against the pinned metadata-extractor 2.21.0.

dschmidt:
My knowledge is limited, but in general it sounds great. As usual we/you need to carefully decide what we want to maintain in tika or fix upstream in metadata-extractor.. no hard opinion on this, I've obviously added functionality to tika myself that could have been added to metadata-extractor

tballison:
Nice! Thank you!

PR #3008, opened 2026-08-11
TIKA-4804: send the terminal signal in fetchAndParseServerSideStreaming

dpol1:
The server-streaming handler never calls onCompleted(), so clients keep waiting after the reply. One line to match the unary handler, plus a regression test. Details and repro: [TIKA-4804](https://issues.apache.org/jira/browse/TIKA-4804).


tballison:
Thank you @dpol1 !

PR #3026, opened 2026-08-14
TIKA-4815: trivial follow-ups from the PR #3022 review

tballison:
What's in it:
  - PipesClientInterruptTest — setStartupTimeoutMillis(1000) bounds the non-interruptible handshake read that made the backoff test's flake window 60 s; both
  tests now try-with-resources the client; @Timeout 30→45 (above the 35 s sum of awaits).
  - TikaGrpcConcurrencyTest — both burst tests now assert pipesParser.isSharedMode() matches the config, so a useSharedServer binding regression is catchable.
  - PerClientServerManager — log no longer says "mid-request" (it also fires mid-handshake) or claims an immediate recycle.
  - grpc index.adoc — Concurrency section ported from the README.


tballison:
@dpol1 wdyt?

dpol1:
the isSharedMode pin is the right call - the burst alone can't prove the mode - and bounding the handshake read kills the one real flake window. LGTM

PR #3088, opened 2026-08-28
[TIKA-4849] Unnecessary dependencies to jaxb-runtime removed

theit:
This PR fixes [TIKA-4849: Unnecessary dependencies to jaxb-runtime](https://issues.apache.org/jira/browse/TIKA-4849):

Two parser modules (tika-parser-miscoffice-module and tika-parser-pdf-module) still contain a depenency to jaxb-runtime although this isn't needed (anymore).

I verified that the build still works by executing "mvn clean verify" on my fork.


THausherr:
I'll run a local test with docker just to be sure.

THausherr:
I'm having troubles building but I think that's another problem which I'll research separately.

PR #3099, opened 2026-08-29
TIKA-4861: Detect raw camera formats by content

dschmidt:
Raw camera formats were detected by extension only; without a name they came back as `image/tiff` and the previews of TIKA-4824 never appeared.

**`RawTiffDetector`** (tika-parser-image-module) tells the TIFF-based formats apart by their image directory:
1. `DNGVersion` tag: DNG
2. vendor `Compression` code (ExifTool's rule): Nikon, Sony, Pentax, Samsung
3. sensor data (CFA/LinearRaw, or a full-resolution single deep sample without PhotometricInterpretation) plus `Make`: the uncompressed variants

A TIFF a camera wrote has RGB data and stays a TIFF. The third rule's shape-based part exists only for the Samsung NX1 and is the weakest; dropping it costs just the NX1 without a file name.

It reads the stream on demand up to 1 MiB (directories sit behind the embedded previews, up to ~460 KB in practice), a plain TIFF costs a few KB, and the stream is reset.

**Also:** magic entries for RAF, RW2, MRW and the other ORF byte orders; `image/x-raw-samsung` added and handled by `RawTiffParser`.

**Verified** without a file name on 21 files from raw.pixls.us (NEF D1 to Z 6, NRW, ARW old and new, SR2, SRW NX100 and NX1, PEF, DNG from Ricoh, Leica and Nokia, RAF, RW2, MRW, ORF, and a CR2 that keeps its type), 20 to 600 microseconds each. Not added to the repo because of their size; the unit tests build their own TIFFs instead.

https://issues.apache.org/jira/browse/TIKA-4861


dschmidt:
Thanks @THausherr for helping out :)

dschmidt:
Wooop wooop 🎉

PR #3107, opened 2026-08-31
TIKA-4864: resolve the default plugins dir against the install layout and make it absolute

dschmidt:
Fixes https://issues.apache.org/jira/browse/TIKA-4864

In the docker image (and any unpacked distribution) TikaServerProcess is loaded from lib/tika-server-core-4.0.0.jar, so the code-source probe looks for lib/plugins, which does not exist. The fallback then returns the relative literal "plugins", which the forked PipesServer resolves against its own working directory. As soon as an orchestrator starts the container with a different workdir (docker run -w, CI systems like Woodpecker do this for service containers), the plugin registry comes up empty and every parse fails with "Unknown fetcher type: 'file-system-fetcher' ... Available types: []".

The fix keeps the code-source heuristic but also probes the parent of the jar directory (covers the lib/ layout) and always returns an absolute path, so parent and forked process agree on the same directory regardless of CWD.

Note: the CWD fallback is an untrusted-search-path pattern (CWE-427 class): the server loads and instantiates jars from whatever directory it happens to be started in. This change narrows the exposure by preferring the install directory and absolutizing, but keeps the fallback since it is documented behavior. An explicit override (CLI flag or env var) might be a good follow-up.

Adds unit tests for all three probe locations plus the absolute-path guarantee.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,10 @@
 Release 4.1.0 - unreleased
 
+   * The default plugins directory is resolved against the install layout
+     (next to the jar, or next to its lib/ directory) and always as an
+     absolute path, shared by tika-server, PipesForkParser and the async
+     CLI; it no longer depends on the working directory (TIKA-4864).
+
    * tika-server error bodies (the 422/500 exception mapper, /meta/{field})
      now honor the exception-reporting policy; /meta/{field} returns the
      already-formatted container exception instead of re-wrapping it with
--- a/tika-pipes/tika-async-cli/src/main/java/org/apache/tika/async/cli/PluginsWriter.java
+++ b/tika-pipes/tika-async-cli/src/main/java/org/apache/tika/async/cli/PluginsWriter.java
@@ -28,6 +28,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 
 import org.apache.tika.config.loader.TikaObjectMapperFactory;
+import org.apache.tika.pipes.core.config.DefaultPluginsDir;
 import org.apache.tika.sax.BasicContentHandlerFactory;
 import org.apache.tika.utils.StringUtils;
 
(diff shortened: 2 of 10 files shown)


tballison:
I _think_ I now get why you'd want this. There are actually four places where this should be fixed.
Could you hoist your `resolveDefaultPluginsDir(codeSourceDir, cwd)` into tika-pipes-core next to ConfigMerger, maybe? WDYT?

| Module / class | Current default resolution | Bug | Action |
|---|---|---|---|
| `TikaServerProcess` (tika-server-core) | jar dir only, then CWD-relative | misses `lib/` layout (TIKA-4864) | fixed by PR #3107; route through shared helper |
| `PipesForkParser` (tika-pipes-fork-parser) | jar dir only, then CWD-relative | same miss of `lib/` layout | replace copy with shared helper |
| `TikaAsyncCLI` (tika-async-cli) | parent of jar dir only, then CWD-relative | misses jar-adjacent layout; also used by `PluginsWriter` | replace copy with shared helper |
| `TikaGrpcServerImpl` (tika-grpc) | none — requires config/`--plugin-roots`; on missing plugin-roots falls back to bare pf4j `DefaultPluginManager` (CWD-relative) | fallback contradicts its own WARN text ("next to tika-grpc.jar") | optional: use shared helper in the fallback |

dschmidt:
I mainly want to make sure the tika image comes up in Woodpecker CI without specifying an explicit working directory. 
Obviously it can be worked around easily, but I'd like it to work out of the box :) 

Checking...

dschmidt:
Done: `DefaultPluginsDir` sits in tika-pipes-core next to ConfigMerger, with the probe order (jar dir, its parent for the lib/ layout, cwd) and the absolute-path guarantee in one place, plus the unit tests moved there. TikaServerProcess, PipesForkParser and TikaAsyncCLI/PluginsWriter route through it; their copies are gone.

I left tika-grpc out: it has no dependency on tika-pipes-core, and its fallback only runs after a TikaConfigException from a config that named plugin-roots, where guessing a directory the config did not name seems wrong anyway. If you want it covered, the helper could move one level down into tika-plugins-core (grpc already uses TikaPluginManager); happy to do that instead.

tballison:
LOL, my agent also noted we should probably change this comment...

```
  Bonus confirmation while in there — the Dockerfile already knew about this bug and papered over it:

  # WORKDIR sets the CWD so tika-server's plugin-root fallback resolves
  # `plugins/` relative to /opt/tika-server (its `getCodeSource()` returns a
  # lib/* path, not the top-level jar, so the "next-to-jar" resolution misses).
  WORKDIR /opt/tika-server

```

PR #3142, opened 2026-09-08
TIKA-4882: Return default value if the extension is unknown in FilenameUtils

timgrein:
Currently the behavior of `FilenameUtils.calculateExtension` only specifies that the `defaultValue` is returned on a parse exception or a null value. The behavior for an unknown value is not specified and silently returns `.bin`. I adjusted the docs and now the `defaultValue` is returned, if the extension is unknown. Also added a corresponding test.

### DIFF
--- a/tika-core/src/main/java/org/apache/tika/io/FilenameUtils.java
+++ b/tika-core/src/main/java/org/apache/tika/io/FilenameUtils.java
@@ -386,7 +386,7 @@ public class FilenameUtils {
      * On parse exception, unknown value or null value, return the default value.
      *
      * @param metadata
-     * @param defaultValue
+     * @param defaultValue, which should include a leading "."
      * @return the extension based on the mime type, including the initial "."
      */
     public static String calculateExtension(Metadata metadata, String defaultValue) {
--- a/tika-core/src/main/java/org/apache/tika/io/FilenameUtils.java
+++ b/tika-core/src/main/java/org/apache/tika/io/FilenameUtils.java
@@ -383,7 +383,7 @@ public class FilenameUtils {
 
     /**
      * Calculate the extension based on the {@link HttpHeaders#CONTENT_TYPE} value.
-     * On parse exception or null value, return the default value.
+     * On parse exception, unknown value or null value, return the default value.
      *
      * @param metadata
      * @param defaultValue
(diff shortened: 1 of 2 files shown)


tika-core/src/main/java/org/apache/tika/io/FilenameUtils.java
```diff
@@ -400,7 +400,7 @@ public static String calculateExtension(Metadata metadata, String defaultValue)
         if (ext != null) {
             return ext;
         }
-        return ".bin";
+        return defaultValue;
```
timgrein, line 403: This could be `null` though. Not sure how that should be handled? We could add `@NotNull` to `defaultValue`, iff that should be the contract

tballison:
Thank you for opening this!

In looking at the codebase, I realize that we let duplicative call sites arise, which we shouldn't have.

EmbeddedDocumentUtil.getExtensionForMediaType(String) already basically does what FileNameUtils does, with some difference input and default missing value.

We should put this functionality in one, along the lines of: 

```
  public static String calculateExtension(Metadata metadata, String defaultValue) {
      String ext = EmbeddedDocumentUtil.getExtensionForMediaType(
              metadata.get(HttpHeaders.CONTENT_TYPE));
      return StringUtils.isBlank(ext) ? defaultValue : ext;
  }

```

Perhaps on a follow on PR, put `public String getExtension(String name) {}` in MimeTypes and then call from there?

Consider deprecating the two current call sites that are nearly duplicative?

timgrein:
@tballison Thanks for taking a look! Happy to take the follow-up, should we merge this PR then now as is and I open a new Jira issue (and the follow-up PR)?

tika-core/src/main/java/org/apache/tika/io/FilenameUtils.java
tballison: I think it should be.

tballison:
Github status looks green, but these are just not running. I wonder if the ASF hasn't paid their bill? :rofl:

timgrein:
> Github status looks green, but these are just not running. I wonder if the ASF hasn't paid their bill? 🤣

image

Seems like they ran now 😄

tballison:
Thank you @timgrein !

PR #3146, opened 2026-09-08
TIKA-4856: maxRenderedPages renders the first pages and reads the whole document

dschmidt:
Rendering only the first page of a PDF used to require `maxPages: 1`, which cuts the text off with it. `"pdf-parser": {"maxRenderedPages": N}` bounds the rendering on its own, counted from the first page, for both `RENDER_PAGES_BEFORE_PARSE` and `RENDER_PAGES_AT_PAGE_END`: a thumbnail request renders one page while text extraction covers the whole document. Same shape as `maxPages`: the first N, not a range, so the two read alike.

Extracted from #3096, where it was the one piece that is pure parser config; the rest of that PR is superseded by the presets from #3118, and I'll close it.

https://issues.apache.org/jira/browse/TIKA-4856


tballison:
From my :robot:, let me know what you think.

```
 Bug (must fix): RENDER_PAGES_BEFORE_PARSE with maxRenderedPages above the page count renders nothing.
  PDFParser.renderPDF now sends PageRangeRequest(1, N), and PDFBoxRenderer.renderRange loops 1..N without clamping to getNumberOfPages(). Page nPages+1
  throws IndexOutOfBoundsException, which is a RuntimeException so it passes the per-page IOException catch, the renderer closes its results, and
  renderPagesBeforeParse records it as a warning and returns with zero renderings. Verified with a probe test on the 2-page fixture, maxRenderedPages: 5:

  RENDER_PAGES_BEFORE_PARSE: renderings=0
    tk:exception:warn=java.lang.IndexOutOfBoundsException: 1-based index out of bounds: 3
  RENDER_PAGES_AT_PAGE_END:  renderings=2  (fine; per-page requests never exceed the doc)

  So the thumbnail case (maxRenderedPages: 1) works, but any N ≥ 2 silently loses every rendering on documents shorter than N. No temp-file leak: the
  renderer's catch (Throwable) closes the partial results.

  Suggested fix: clamp to in PDFBoxRenderer.processRequest (Math.min(to, pdDocument.getNumberOfPages())). That matches PopplerRenderer, since pdftoppm -f 1
  -l 5 on the same fixture writes exactly 2 PNGs. Keep from beyond the last page throwing, because PDFBoxRendererTest.testFailedRenderLeavesNoTempFiles uses
  PageRangeRequest(9999, 9999) as its vehicle. Add the failing test: both strategies, limit above the page count, assert nPages renderings. The PR's tests
  only cover maxRenderedPages: 1 on a 2-page doc, which can't reach this.

  Hygiene
  - tika-parsers-standard-integration-tests/.../config-examples/pdf-parser-full.json is the "every knob" example; add "maxRenderedPages": -1 next to
    maxPages with a one-line comment.
  - The PR description's "used to require maxPages: 1" is only true for AT_PAGE_END; BEFORE_PARSE never honored maxPages at all. Wording only.
```

PR #3148, opened 2026-09-08
TIKA-4856: thumbnails catalog preset and ThumbnailUnpackSelector

tballison:

* the issue ID (`TIKA-XXXX`)


tballison:
@dschmidt This is 100% :robot: Let me know what you think. I need to spend some time reviewing this and figuring out if there's a cleaner option.

dschmidt:
High level this fits both of my use cases, but it ships only one of them.

The "just give me the thumbnail" case is covered by this PR as is: the selector keeps exactly one file in the zip, so my client doesn't need any picking logic anymore, and NO_OCR + skipOcr + maxPages 1 keep the parse cheap. PUT /unpack/preset/thumbnails, done. That fully replaces my endpoint from #3096.

My other use case is indexing: metadata, full text and all embedded files in one request, with OCR left on, but previews still rendered (PDF page 1, EMF/WMF) so I get dimensions etc. That needs a second catalog preset: no selector (I want all embedded files), no NO_OCR, and crucially maxRenderedPages instead of maxPages, because cutting the text off at page 1 is exactly what I can't have there. Something like:

```json
 "render-thumbnails": {
    "pdf-parser": { "imageStrategy": "RENDER_PAGES_AT_PAGE_END", "maxRenderedPages": 1,
                   "ocr": { "dpi": 96, "imageType": "RGB" } },
   "emf-parser": { "renderImage": true, "renderOnlyEmbeddedResourceTypes": ["THUMBNAIL"] },
   "wmf-parser": { "renderImage": true, "renderOnlyEmbeddedResourceTypes": ["THUMBNAIL"] }
 }
```

That's also the reason for #3146: the thumbnails preset here can arguably keep maxPages 1 (cheaper, and /unpack discards the text anyway), but the indexing preset can't work without maxRenderedPages.

tballison:
Y, I'm worried about the alignment between the two calls. With the selector, are we now skipping embedded files, and will the ids be identical between the "full" call and the "thumbnail" call. I _think_ so because the unpack selector chooses only the bytes for unpacking, it shouldn't affect the overall parse?

Would it be better to do the full extraction to get /rmeta, and then have something in the unpack endpoint that allows the extraction of one or more attachments by id number?

I'm sorry for the churn on this. Short of storing all the images, this is actually kind of challenging. Thank you for working through this as a general solution.

tballison:
Ugh. No, that's even worse across versions of Tika. If you store thumbnail is at id=3, and then we change parsing in a later version, that will break. Alignment for your use case doesn't matter. You just want the the first thumbnail looking thing on the second parse, right?

dschmidt:
No worries, the whole topic is a bit confusing and tough to crack. Thanks for your effort in solving my request in a sustainable matter. Happy to help with that wherever I can. 

Right now I'm completely overtired and can't think straight tho 😴

So I'm afraid I can't fully answer your questions right now, but yeah, I can define my requirement or my initial issue further: 

I found it harder than it should be to figure out the right request options to get all possible thumbnails. Also it looked like it would be subject to subtle changes from tika version to tika version. 

What I need in the end is a response for which I can implement a simple rule to get the thumbnail - in the two scenarios described above. 
Something like first thumbnail or thumbnail which is a rendering is easy enough to implement. 

If we can reduce the size of the response, that would be nice but not a requirement for me. 

To make it more concrete: I want to implement thumbnails in OpenCloud without having to adjust the rules specifically for every tika version and I don't want to know which formats tika provides thumbnails for at all. 
That's why I need/want a generic form which is stable

tballison:
Should have a clear mind for this next week. Apologies for the delay!

dschmidt:
No worries, you don't have to work on this, so I'm grateful anyhow

tballison:
Superseded by: https://github.com/apache/tika/pull/3179

PR #3179, opened 2026-09-15
TIKA-4856: thumbnails catalog preset and ThumbnailUnpackSelector

tballison:

* the issue ID (`TIKA-XXXX`)


docs/modules/ROOT/pages/using-tika/server/index.adoc
```diff
@@ -154,6 +154,39 @@ not (detectors, embedded-document extraction, exception reporting, ...) and rais
...
+logs one line per active preset naming the components it configures.
+
+==== The `thumbnails` catalog preset
+
+Tika ships one catalog preset, `thumbnails` (in `tika-pipes-core`), for
+`/unpack/preset/thumbnails`. Nothing is active until the operator names it:
+
+[source,json]
+----
+{ "presets": { "thumbnails": true } }
```
dschmidt: What is this? Config? Request body? ...? 

If config, does this mean the preset is not available by default?

dschmidt:
overall looks great (afaict by looking at it on my phone screen), it's just missing a preset that enables rendering without disabling ocr for the "give me everything use case"

docs/modules/ROOT/pages/using-tika/server/index.adoc
tballison: We should talk about this. I'm really worried about presets going automatically "live" by default. These two are good, but going forward, we have to ensure that devs will not add a default that is dangerous in some environment.

The current PR defines presents, but the operator still has to turn them on.

tballison:
> overall looks great (afaict by looking at it on my phone screen), it's just missing a preset that enables rendering without disabling ocr for the "give me everything use case"

added render-thumbnails. from the :robot: `/rmeta/preset/render-thumbnails reports are the image /unpack/preset/thumbnails returns`

dschmidt:
Great! Now wondering... should it be (render-)thumbnails oder thumbnail? We use it for the main thumbnail of the original file, so it's only ever a single one, no?

docs/modules/ROOT/pages/using-tika/server/index.adoc
tballison: My proposal is to go forth with requiring operators to turn presets on as the default. We can discuss on a separate issue whether we want to allow some on by default?

tballison:
> so it's only ever a single one

:face_palm:

dschmidt:
Great! I'm good with this now, I think!

tballison:
Thank you for your patience and collaboration on this one, @dschmidt !

dschmidt:
Likewise! ☺️

Any plans for a 4.1 release anytime soon?

dschmidt:
oh and... the preset is not enabled by default?

tballison:
Sorry, it isn't. See: https://github.com/apache/tika/pull/3179#discussion_r4025164253

docs/modules/ROOT/pages/using-tika/server/index.adoc
dschmidt: Ah, sorry, missed this. 

Will you open the ticket? 
If it's not the default I would at least like to make it very easy to opt in, maybe via env var or so? 
Creating a config file is (comparatively)  pretty cumbersome in tutorials, compose setups etc

